PR #20, opened 2016-11-01
Resolves issue 6  and adds author/date to blog posts

ostermanj:
Should resolve issue 6. Sidebar filters recent blogs for those other than current. Also adds author and date to individual blog posts.

### DIFF
--- a/docs/_layouts/blog_post.html
+++ b/docs/_layouts/blog_post.html
@@ -23,31 +23,34 @@
 	{% include navbar.html %}
 		<!--main body-->
 		<div class="container">
+            <div class="row">
+                <div>
+                    <h1>{{ page.title }}</h1><p>{{ page.author }} | {{ page.date | date: '%B %d, %Y' }}</p>
+                </div>
+            </div>
 			<div class="row">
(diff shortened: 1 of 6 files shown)


docs/_layouts/blog_post.html
```diff
@@ -23,31 +23,34 @@
 	{% include navbar.html %}
 		<!--main body-->
 		<div class="container">
+            <div class="row">
+                <div>
```
NealHumphrey: Just a heads up, when using row in bootstrap, you have to have at least one nested col class (col-xs-12), even if it is just a 12 width column. Column adds a positive margin and row adds a negative one. Currently the title and body are misaligned b/c of this. I'm merging the branch and will just do a quick fix rather than kick this back but just wanted to flag it for you!

NealHumphrey:
This all looks good, tested it locally and it works. Thanks John! Hope to see you there tonight too.

docs/_layouts/blog_post.html
ostermanj: Thanks!

PR #25, opened 2016-11-04
Add DC map visual

DanHenebery:


NealHumphrey:
Instead of merging into `dev`, I merged this directly into a new branch: 22-d3-land-value . (22 is the issue number that this feature relates to, and land-value is the visualization we are trying to create)

NealHumphrey:
Hey @DanHenebery, I did a quick rearrangement of the files you added, and inserted the necessary dependencies to index.html - got your visualization loading within the Housing Insights page. Let me know if you think you'll be able to take a look at reconfiguring for this use. For the currently envisioned setup the bars themselves won't change since the bar chart will also be neighborhood clusters - which differs from your current data structure, but it might not need significant restructuring if for now we just use the 'Washington, DC' location name.

PR #29, opened 2016-11-04
22 highcharts landvalue -- adds bar chart, links hover states

ostermanj:
Ok. This is working pretty nicely. Needed to render charts to different containers; looks like one Highmaps constructor can render a map alongside a pie chart, but other chart types need a separate Highchart constructor.


NealHumphrey:
Cool - I'll check this out soon. In case you want to see what it looks like with real data, I have (un-checked!) data in the 22-SQL-landvalues branch, which I just pushed. I did both total value and also per-affordable-apartment stats. Interesting stuff!

NealHumphrey:
And by soon, probably after class tomorrow - gotta take off for now

ostermanj:
cool. I've connected it a new branch, displaying sum_appraised_value_current_total from summary_tax_values.csv. making new pull request.

PR #30, opened 2016-11-04
22 sql landvalue

NealHumphrey:
@froi If you have a chance between now and Sunday morning (no worries if you won't be able to!), could you check my work on SQL? I'd like to get another set of eyes on it if I can to make sure the data is at least close to accurate. 

I started working from the code you wrote last night and spent a while debugging and looking for both data quality issues and understanding the one-to-many relationships. A couple things came from me not yet understanding the data in the guidance I gave you, and from digging deeper. 

I found:
- Turns out multiple projects can have the same SSL
- real_property has multiple records per SSL - it records data of transactions. So we shouldn't use this as a mapping to tax. Parcel is one-to-one.

I think I fixed both of those in my current query. 

Also:
-Some SSLs aren't in the tax database (so sums will undercount)

I started doing the necessary counts to interpolate this, but have not yet actually done calculations on this.


NealHumphrey:
@froi - if you don't think you'll have time to take a look just let me know too

froi:
Sorry I didn't answer until now. I haven't been able to sit down to look at this. I am now though. I'll let you know if I find anything.

froi:
I think you're on the money here. I think we should have a data definition/model somewhere. Maybe the readme should be updated to explain the data a bit?

NealHumphrey:
@froi  thanks for the QA! Agree on data dictionary as next step. I'll start one in `docs/resources` and we can be sure to add any other insights we find there. All the data sources I loaded into SQL include data dictionaries with short descriptions of each column name, but they are not always informative and didn't have clear info like how exactly the one-to-many relationships work, so some we'll just have to discover via experimentation.

PR #31, opened 2016-11-05
22 connects real data

ostermanj:
Connect highchart map and bar chart to summary_tax_values.csv


NealHumphrey:
Cool! This looks great. I'll add some intro text and take a stab at adding the neighborhood labels to the left of the bar chart. We should have a great demo to show people on Tuesday. Thanks @ostermanj, and also @froi for helping generate the data.

PR #41, opened 2016-12-07
Added grocery store location data

salomoneb:
Pulled grocery store data from http://opendata.dc.gov/datasets/1d7c9d0e3aac49c1aa88d377a3bae430_4. Thought this might be worth looking at.


NealHumphrey:
Good stuff, merging in. Do note, we were talking with Kate Rabinowitz from Code for DC, who works at OCTO, and she said be aware it is likely not up to date. We'll use for now, though, just flagging it.

PR #55, opened 2017-01-06
Adds the Building Permits data set to the manifest, and fixes a couple loading issues

NealHumphrey:
Building permits data **has** been loaded into the AWS database. No need to re-run

Solves:
- Low Memory error that showed up in the read_csv method
- Non-printing characters at the start of the file (i.e. in the header name of the first column) now removed


NealHumphrey:
@salomoneb This fixes the building permit issue. You should pull the latest dev branch, but no need to run the script - I already ran it for the 2016 opendata.dc.gov data you pointed to.

PR #56, opened 2017-01-09
Added several queries and output_data csv files 

NealHumphrey:
that are needed for the Blocks prototype.

In addition, the Table_Shells file is a reference file for all the data that is available through American Community Survey in a (kind of) easier to navigate document.


NealHumphrey:
@salomoneb check it out, I added a few useful queries joining some of our data to get the prototype team some answers they need. Useful for our exploration. All queries saved to the /scripts/analysis folder, and the created files to output_data

PR #57, opened 2017-01-11
D3 scaffolding

ostermanj:
moves the d3 scaffolding from jsfiddle to a repo


NealHumphrey:
Hey john, as we discussed we should move this into the /docs/static/js folder (or something similar) in the near term. Approving this merge as-is, though - can refactor from this version since we don't have any live code right now.

Issue #54, opened 2017-01-06
Refactor CSV import code to be more efficient


NealHumphrey:
------------------------
update

This has been partially completed and is in the overhaul-csv-loading branch. 

The remaining tasks that need to be completed are 
#83 
#86

This ticket should be closed, but keeping in the 'on hold' column so it is available for reference if needed. 

----------------------------------------------------
## 4 parallel tasks that we can do now:
- [x] `(salomone)` Make JSON data: We will store our list of expected fields and their data types into a file called meta.json. Use Pandas to take a new .csv file, guess the data types and produce the meta.json. Use the `make_draft_json.py` file as a starting point. 

- [x] `(Brandon)` Figure out how to read a CSV into memory (so that we can edit values such as messy data) and then write it to SQL efficiently. Use `demo_load_to_sql_without_pandas.py` as a starting point. 

- [ ] `(unclaimed)` Everything else for our `load_data.py` file, which handles downloading files if needed, comparing manifest.csv to the manifest table in SQL to decide whether to load a table or not, handling errors. Temporarily just use dataframe = pandas.read_csv and dataframe.to_sql() in the csv_to_sql() function until #2 is ready. Use my list of functions and the rough outline in the main() function as a guide, but there are probably additional things I didn't think of. 

- [x] `(Neal)` Get data upload tools to S3 configured. 

Stuff to do after:
- [ ] run make_draft_json.py on all our data
- [ ] set up data cleaning phase in between read and write (see note in the `csv_to_sql()` function)

To get working:
- Pull the branch `overhaul-csv-loading` into your repo
- Make a new branch off of that branch for your portion
- When you're ready to share your work, make a pull request on Github into the `overhaul-csv-loading` branch (*not* the dev branch). We'll consolidate all our work there and make sure it's working before merging into dev. 
(let me know if you need help with any of this!)

## *What you'll see in the /scripts folder:*
**/`__root__`**
- database_management.py: contains tools to connect to our database. Moved to root because this can be also be used later in other places (e.g. analysis scripts). I created this for housing_risk but I have not yet tested this in our project, so it might need minor changes. 
-secrets.example.json: check out the addition of the local_database vs. remote_database, and make this edit to your secrets.json copy (or ask me for a copy if you don't have one yet).

**/data**
- I moved the small_data, big_data, and output_data folders into here to clean up our root

**/ingestion:**
- make_draft_json.py: for item #1
- demo_load_to....: for item #2
- load_data.py: for item #3
[...]

PR #61, opened 2017-01-26
Added JSON data loading script

salomoneb:
Removed previous pull request and re-created branch because things got messy. 

This script produces a JSON file per the first item in #54.

### DIFF
--- a/scripts/ingestion/make_draft_json.py
+++ b/scripts/ingestion/make_draft_json.py
@@ -5,12 +5,6 @@
 '''
 Loads our flat file data into the Postgres database
 '''
-
-
-##########################################################################
-## Imports & Configuration
-##########################################################################
-#external imports
 import logging
 import json
 import pandas as pandas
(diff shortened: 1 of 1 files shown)


NealHumphrey:
Hey @salomoneb  this looks great! Everything works well and is well set up - but we do need to figure out how to resolve that data type issue. That's not something I had fully anticipated / thought through. I'm not familiar with the numpy.dtype object and the documentation didn't immediately have a clean explanation of what the possible options are. It looks like we should dig into the documentation on the numpy.dtype class](https://docs.scipy.org/doc/numpy/reference/generated/numpy.dtype.html#numpy.dtype) and maybe (?) the [numpy.dtype.descr attribute](https://docs.scipy.org/doc/numpy/reference/generated/numpy.dtype.descr.html#numpy.dtype.descr). Or, google around on how to better access a list of possible numpy.dtype options especially those parsed by Pandas.read_csv....

NealHumphrey:
Hey @salomoneb FYI I'm merging this into the overhaul-csv-loading branch despite the outstanding data type issue - that way we can all work from the same stuff even though it's still a WIP. You should pull the latest overhaul-csv-loading and make a new branch from there.

salomoneb:
Pulled the latest overhaul-csv-loading. I was hoping that the numpy -> postgres data type mapping would be an easy fix, but from some preliminary searching, it looks like we may need to develop our own solution. I'm going to continue look and/or come up with something in the next few days.

PR #63, opened 2017-01-28
included basic slider viz work, not yet integrated into Jekyll

wassona:


ostermanj:
Merged into Dev so we can all see how it works. Thanks!!

PR #64, opened 2017-01-31
Re: issue 52 - creating constructors that inherit from Chart 

ptgott:
(I accidentally created this in my own repo yesterday--still figuring github out!)

I was going to take a stab at making a donut chart, then thought it might be useful first to make part of the Chart constructor reusable for different kinds of charts. I've extracted a separate constructor for the moving block chart and added some basic inheritance (all in prototype-1a.js).


ostermanj:
That sounds good. Thanks, @ptgott .  Can review and resolve conflicts at tonight's work session. Will you be there? I get the gist of what you've done but would have some questions.

ptgott:
Sounds good. I'll be there!

ostermanj:
Merged 52-chart-inheritance.

PR #70, opened 2017-02-06
Overhaul csv loading

blw9u2012:
* Add DataReader which will read each csv file row by row. 
* Add DataWriter which will maintain and write rows to dirty and clean csv files from the cleaned source.
* Add a Cleaner which will clean the csv files based on known type. 
* Began to create a Postgres utility file which will handle the interaction with postgres.


NealHumphrey:
Merging as-is. I will delete the /mockups/d3 stuff post-merge, and will leave the /scripts folder in place until I hear back from @blw9u2012 re: 'correct' version of the /scripts/demo_load....  vs. /python/demo_load.... files.

PR #71, opened 2017-02-07
merge with base

wassona:


wassona:
Was trying to update my fork and got head and base reversed. Oops.

PR #72, opened 2017-02-08
Added subdirectory to contain a prototype for the single-building view.

vincecampanale:


vincecampanale:
Definitely - changing it now.

PR #75, opened 2017-02-09
Provide baseline javascript for all Charts;Recreates moving blocks in…

ostermanj:
… new file structure

@NealHumphrey: Branch was protected from direct push, so I'm making a pull request

Description:
Takes the scripts necessary for all Charts and puts that into housing-insights.js. Removes the IIFE structure so methods and variables are accessible across files. All test projects should link to this file before their own chart-specific scripts.

Housing-insights.js handles the initializing and initial setup of Charts, with the option of fetching new data or using data that another Chart has already fetched. Uses a pub/sub module to publish when certain data is loaded so that subsequent, subscribe constructors can fire only when that data is ready.

This PR also recreate the moving-block prototype in moving-blocks.html. The scripts in moving-blocks.js should serve as an example of how to inherit from Chart in housing-insights.js, how to use the pub/sub, and how to call new constructors.

### DIFF
--- a/javascript/tool/building.html
+++ b/javascript/tool/building.html
@@ -5,6 +5,8 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Building View</title>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/d3/4.4.1/d3.min.js"></script>
+  <link rel="stylesheet" type="text/css" href="css/tool.css">
 </head>
 <body>
   <header class="building-header">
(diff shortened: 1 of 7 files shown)


NealHumphrey:
@ostermanj

Yes - as I mentioned yesterday, I'm rearranging some of the git permissions both to accomodate waffle.io as well as make sure once we start having breakable code (soon!) we make sure to always go through pull requests so there's a second pair of eyes on everything. And calls everyone's attention to the change so they can update their branches and see the written comment of intent. 

For now I'll assume this is working - from a scan it looks reasonable. And there's nothing we can break yet :)

ostermanj:
Yeah, totally get having it reviewed. Thanks!

Issue #50, opened 2017-01-03
Zoning / PUD application status nearby


NealHumphrey:
We'd like to include notifications of nearby PUD / Zoning statuses in a map + list for each PresCat building. 

We will probably want to talk to the Office of Zoning about getting access to this data systematically. They have an online system and posted PDFs. For now, this is the information available:

[PDF list of filings by Ward](http://planning.dc.gov/page/zoning-commission-cases-zc)
[Interactive Zoning Information System](https://app.dcoz.dc.gov/Help/Home.html)
[Board of Zoning calendar of upcoming hearings](https://app.dcoz.dc.gov/Content/Schedule/Calendar.aspx)
[Zoning Commision calendar of upcoming hearings](https://app.dcoz.dc.gov/Content/Schedule/ZCCalendar.aspx)
[Zoning map](http://maps.dcoz.dc.gov/zr16/#l=16&x=-8571256.999999998&y=4708437.499999998&mms=18!26!21!24!22!19!4!8!1!2&dcb=0) (contains current pending PUD areas in red)

salomoneb:
I'm still getting errors when I try to add to the database, but here's a CSV of (I think all of) the zoning data on http://maps.dcoz.dc.gov/zr16/. [https://www.dropbox.com/s/5ntjt1pxgwy58ul/pud.csv?dl=0](https://www.dropbox.com/s/5ntjt1pxgwy58ul/pud.csv?dl=0)

The code that I used to generate it is here: https://gist.github.com/salomoneb/4eaec0e1f578418eb682151bb2e6816f

I scraped the data by querying an endpoint that I found on the map page: `https://maps2.dcgis.dc.gov/dcgis/rest/services/DCOZ/Zone_Mapservice/MapServer/25/query?f=json&where=Square ='` + SQUARE NUMBER + `'&returnGeometry=false&spatialRel=esriSpatialRelIntersects&outFields=*`

The script constructs URLs using each property square number in the `dc_tax` table and requests data from each one, outputting the results in the CSV. 

This is the first real Python script I've written and wildly impractical for ongoing purposes, but it does give us a large baseline snapshot. We'll also probably need to clean the data because many of the values seem to be in inconsistent formats. 

If you have a contact at the Office of Zoning, getting direct access to the database is probably the way to go.

salomoneb:
One thing to point out now that I'm looking at this on my work computer with a functional copy of Excel: check out the "case_type_relief" column, which has at least six PUD-related categories.

[...]

PR #79, opened 2017-02-11
Added script to scrape zoning data

salomoneb:


salomoneb:
- Scraped CSVs are in `data/raw/zoning_data/full_separated_data`.
- Data without a `case_type_relief` value are in `empty_zoning_data.csv`.
- Data with that value are in `existing_zoning_data.csv`.

salomoneb:
Now that we have answers to some of our questions (notes documented [here](https://github.com/codefordc/housing-insights/issues/50)), what needs to be done for this to be merged? I still intend to follow-up with the zoning coordinator about which data is most accurate/inaccurate. Anything else for now?

NealHumphrey:
@salomoneb I think it should be good to merge - maybe we can just take a quick look at it together tonight and make sure that our zoning conversation doesn't change what data we need included on this scrape or if it's just all relevant to how we use it later?

PR #81, opened 2017-02-12
Added a simple Mapbox map to the building page.

ptgott:
Please see my comments in the file building-map.js for ongoing issues. This is still very much a work in progress!

### DIFF
--- a/javascript/tool/building.html
+++ b/javascript/tool/building.html
@@ -12,14 +12,13 @@
   <link href='https://api.mapbox.com/mapbox-gl-js/v0.32.1/mapbox-gl.css' rel='stylesheet' />
   <!-- end Mapbox requests -->
   
-  <!-- begin project-specific asset calls -->
+  <!-- begin project-specific STATIC asset calls.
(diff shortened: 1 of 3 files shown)


ostermanj:
@ptgott this all looks good to me. I have one question: 

You moved the project-specific asset calls up in to the `<head>` which of course makes good sense, especially for the static assets, but we then do have to be sure not to call  constructors, etc, before the DOM element they attach to is ready to be manipulated. You've done that in building-map.js with window.load listener.

Do you think avoiding the issue by calling the script at the end of `<body>` is not worth the somewhat less clean organization? Or is it bad practice in general?

ptgott:
The reasoning for putting the script tags in the body makes sense. I placed them in the head out of personal habit before adding anything else to the html file. Beyond this, I've got no problem moving them to the body. Since we're still starting out on the building page, we can leave the tags in the body and re-visit the issue depending on what we need to call, when we need to call it, what sequence/timing issues exist and so on.

Issue #73, opened 2017-02-08
Calculate metro and bus route distances


NealHumphrey:
Using Mapbox's [Python SDK](https://github.com/mapbox/mapbox-sdk-py), calculate the number of (a) metro stops and (b) bus *routes* within 0.5 miles walking distance of each building in the `project` table from our database. 

- num. of bus routes should be defined as a bus route that stops at a bus stop within the distance parameter. If there are two stops in the radius that stop at a particular
- This ticket can be ready-to-merge into our dev branch if it uses a hard-coded list of addresses/lat-lon pairs, or if it reads directly from the [csv file of projects](https://s3.amazonaws.com/housinginsights/raw/preservation_catalog/20160401/Project.csv) downloaded locally. Other tickets will be developing re-usable modules for reading data from our database. 
- Bus stops should use [WMATA's JSON API](https://developer.wmata.com/docs/services/54763629281d83086473f231/operations/5476362a281d830c946a3d6d)
- It will likely be necessary to call WMATA's API with the 0.5 mile crow-fly distance, and then filter that to any stops that are not actually available within the Mapbox-provided walking distance.

salomoneb:
Very quickly grabbed the location data for [Metro stations](https://developer.wmata.com/docs/services/5476364f031f590f38092507/operations/5476364f031f5909e4fe3311) and [bus stops](https://developer.wmata.com/docs/services/54763629281d83086473f231/operations/5476362a281d830c946a3d6d/console)

[transport_data.zip](https://github.com/codefordc/housing-insights/files/771781/transport_data.zip)


.

NealHumphrey:
Cool @salomoneb ! 

@prisalex came to the Code for DC hacknight and joined up on the project; she started looking into this then (but I'm guessing w/ amount of time available probably didn't get very far). We can coordinate on this tomorrow and see what help she needs.

prisalex:
Hello there, was actually planning on working on this today.   I'll keep you all posted on progress...

salomoneb:
Cool. If you didn't already see it, Mapbox's [Distance API](https://www.mapbox.com/api-documentation/?language=Python#distance) looks like it might be good for this.

prisalex:
[...]

PR #87, opened 2017-02-14
Calculate rail/bus distances to project

prisalex:
This is a fix for #73.

2 notes about the script and output itself. I deleted my personal API keys for Wmata and mapbox (also didnt want those publicized on github) so those need to be added at the top in order for this to run.   I delimited the lines/routes in the info.csv with colons since the number per stop varies.   If we are using a DB that supports json columns, I could change to a json structure and potentially save more data but thought this might be sufficient. 

I wasn't sure exactly where to put the script and the output CSVs so let me know if I should move them. 

This is my first python script so let me know if there are stylistic or coding conventions I should be following.


scripts/wmata_data/distcalc.py
```diff
@@ -0,0 +1,149 @@
...
+    infoCsvWriter = csv.writer(infoOutputFile)
+
+    reader = csv.DictReader(projectsFile)
+
+    numrow = 0
+
+    #print headers
+    distCsvWriter.writerow(('Nlihc_id','type','stop_id_or_station_code','dist_in_miles'))
+    infoCsvWriter.writerow(('code_or_id','type','name','lat','lon','lines'))
+
```
NealHumphrey: It looks like the infoCsvWriter does not have the same data as it's headers indicate. Maybe you refactored to remove 'type' from the data list? if you open wmatainfo.csv 'name' column has 'lat' in it.

scripts/wmata_data/distcalc.py
```diff
@@ -0,0 +1,149 @@
...
+def writeRailInfo(railStations, infoCsvWriter):
+
+    print("Writing RAIL INFO");
+
+    for station in railStations:
+        #delimit list of lines with colon
+        lines = station['LineCode1'] #all stations have at least one line
+
+        #there is probably a more elegant way to write this...
+        if station['LineCode2'] != None:
```
NealHumphrey: A cleaner way - tackle all the cases with the same method. You'll need to initialize `lines` first as empty string.
`for line_code in ["LineCode1", "LineCode2", "LineCode3", "LineCode4"]:`

scripts/wmata_data/distcalc.py
```diff
@@ -0,0 +1,149 @@
...
+            lines = '{}:{}'.format(lines, station['LineCode2'])
```
NealHumphrey: For this context, better way is `lines += station['LineCode2']` or maybe you might need to use `lines += str(station['LineCode2'])` depending whether the parsed json needs explicit string coercion. `a+=b` is equivalent to `a = a + b`

scripts/wmata_data/distcalc.py
```diff
@@ -0,0 +1,149 @@
+import sys
+import csv
+import requests
+
+#TODO replace these with code4dc keys
+WMATA_HEADERS = {'api_key': ''}
+MAPBOX_PARAMS = {'access_token': ''}
+
```
prisalex: I wasn't sure where this script should go so rather than hardcoding the secret.json path in I just added it as an arg to the script.  For now.  If you know where dist.py is going, I can hardcode it although wasn't sure if that is what you all are doing in other scripts.

scripts/wmata_data/distcalc.py
```diff
@@ -0,0 +1,149 @@
...
+def getMiles(i):
+    return i*0.000621371192
+
+###### Converst miles to meters
+def getMeters(i):
+    return i*1609.344
+
+###### Gets walking distance between two locations in meters
+def getWalkingDistance(srcLat, srcLon, destLat, destLon):
+    distReqCoords = srcLon + ',' + srcLat + ';' + destLon + ',' + destLat;
```
prisalex: Done!

scripts/wmata_data/distcalc.py
```diff
@@ -0,0 +1,149 @@
...
+    data = response.json()
+
+    for stop in data['Stops']:
+        walkDist = getWalkingDistance(lat, lon, str(stop['Lat']), str(stop['Lon']))
+
+        if walkDist <= radiusinmeters: #within 0.5 miles walking
+            distCsvWriter.writerow((project['Nlihc_id'],'bus',stop['StopID'],"{0:.2f}".format(getMiles(walkDist))))
+
+
+####### Writes rail station metadata to given CSV writer
```
prisalex: Done!

scripts/wmata_data/distcalc.py
```diff
@@ -0,0 +1,149 @@
...
+def writeRailInfo(railStations, infoCsvWriter):
+
+    print("Writing RAIL INFO");
```
prisalex: Fixed

scripts/wmata_data/distcalc.py
prisalex: Good catch.  I don't know how I didn't notice this originally.  Anyway, fixed now.

scripts/wmata_data/distcalc.py
```diff
@@ -0,0 +1,149 @@
...
+    numrow = 0
+
+    #print headers
+    distCsvWriter.writerow(('Nlihc_id','type','stop_id_or_station_code','dist_in_miles'))
+    infoCsvWriter.writerow(('code_or_id','type','name','lat','lon','lines'))
+
+    railResponse = requests.get("https://api.wmata.com/Rail.svc/json/jStations",headers =WMATA_HEADERS)
+    railStations = railResponse.json()['Stations']
+    writeRailInfo(railStations, infoCsvWriter)
+    writeBusInfo(infoCsvWriter)
```
prisalex: So I refactored a bit keeping in mind the idea that one might want to call a func from the other module.  The main reason I did it differently was because I use the railStations json object later when computing the distances for the projects, so I don't have to make extra API calls.  But I refactored it so it can accomplish both goals of keeping railStations json for later processing & being able to call the func from other modules if desired.

scripts/wmata_data/distcalc.py
```diff
@@ -0,0 +1,149 @@
...
+        findRailStations(railStations,row,radius,distCsvWriter)
+        print("Completed processing rail stations for {}".format(numrow))
+
+
+        # find all bus stops within 0.5 miles
+        print("Starting processing bus stations for {}".format(numrow))
+        findBusStations(row, radius, distCsvWriter)
+        print("Completed processing bus stations for {}".format(numrow))
+
+
```
prisalex: Fixed.

scripts/wmata_data/distcalc.py
prisalex: Fixed

scripts/wmata_data/distcalc.py
prisalex: Fixed

prisalex:
I think I addressed everything.  I'll upload the csvs to S3 shortly...

NealHumphrey:
@prisalex  Cool - let me know if those S3 instructions are complete or not. I think only one other person (salomone) has tried using them since I wrote them. 

 In short, you want a folder called 'data' in the root of the project directory. The contents of the data folder will exactly match the contents of the s3 bucket. You'll want to put your WMATA data into a folder within there that makes sense (probably data/raw/wmata or somethign). There's about a GB of data there now, so if you sync download (you'll probably want to do that eventually) it may take a little time to run the first time, and be aware of what it'll do to your hard drive space if necessary. 

Feel free to call if you have issues / questions

scripts/wmata_data/distcalc.py
NealHumphrey: Sounds like a good intermediate approach. In the refactored branch, `scripts` gets moved down one level, so it will end up in `repodirectory/python/scripts/wmata_data`. Most of the Python code in the refactored approach actually ends up in a package... probably easier to show you those details tonight.

scripts/wmata_data/distcalc.py
NealHumphrey: Sounds good. Eventually we'll want all of our get-data scripts to be called from a cron job somewhere, hence the desire to be able to call from elsewhere. Honestly when we do that, we'll probably want to move everything into the package code anyways (instead of in a script), so it'll need another refactor to accomplish that - but the structure for how we will approach all these get-data functions has not been set up yet, so this will be more than enough for now.

PR #89, opened 2017-02-15
Adds more comments to housing-insights.js and moving-blocks.js to mak…

ostermanj:
…e it easier to follow example. Adds width and height to parameters of specific constructor

Almost all changes are more / better comments hopefully to make it easier to follow the example of inheriting from the shared Chart constructor to make specific charts.

Only substantive change is to add width and height of the SVG to the parameters of the constructor, rather than hard coding it in to the prototype. Much more flexible.

### DIFF
--- a/javascript/tool/js/housing-insights.js
+++ b/javascript/tool/js/housing-insights.js
@@ -1,24 +1,53 @@
 'use strict';
 
+/*
+ *  This file contains the methods and variables common to all d3 charts for housing insights. Specific charts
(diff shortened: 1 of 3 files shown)


ostermanj:
Thanks @ptgott ! I'll merge it now.

BTW, for those following: My comments state that the first parameters of the specific chart constructors should match the parameters of the inherited Chart constructor, but Rich got an example working where his constructor for donut charts leaves out parameters I figured would be necessary, and it still works. So that remains a little unclear.

PR #93, opened 2017-02-19
Adding the beginnings of code for feeding building data into Mapbox maps for the building page

ptgott:
### Something to review
In the building page, I'm working on a way to grab data for Mapbox maps. As shown in the building page prototype, we're displaying various layers of data (probably geoJSON) on Mapbox maps: layers of public transport, other public housing, building permits and so on. I was thinking of using app.getData() to do this, grabbing the .csv files that we use for Charts and manipulating the data into geoJSON (if it's not in that format already). This way, I could add geoJSON layers to Mapbox maps while sharing data with any other Charts that needed it. 

There's a possible issue here: the parameters of app.getData() currently assume that we will be using the data for a Chart. In the comments I've suggested two possibilities, either (a) adjusting app.getData() to work with both Charts and Mapbox objects; or (b) having Mapbox maps make separate requests, pulling geoJSON straight from the server. Let me know which you prefer and I can add to my '80-map-on-building-page' branch.

### A more general possibility
Since different Charts (or Mapbox maps) will be on different pages of the site, we could think about putting the 'app' object in some kind of client-side storage, e.g. [Window.sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage). This way a user could, say, access the moving block chart, then access the building page, reusing the data from the former for the latter without the code having to process the .csv file again.

### DIFF
--- a/javascript/tool/js/building-map.js
+++ b/javascript/tool/js/building-map.js
@@ -1,4 +1,4 @@
-(function(){
+(function prepareBuildingMaps(){
   'use strict'
   
   // FOR ACTION - determine how to bind building data to a building page. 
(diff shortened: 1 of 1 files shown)


ostermanj:
@ptgott I think your session storage idea is right on and wouldn't be hard to implement. The code could check for the object in local storage (if local storage exists) and fetch new data only if it's not there already. I or someone could work on that tonight. It occurs to me, too, that the Chart constructor shouldn't require a developer to know, regardless of local storage or not, whether their chart's data would already have been loaded by another chart. The code should instead check to see if there's already a data object from the requested source and, if so, skip reloading it. That could be another task for tonight. Currently I have us putting in a null value for the source—that's pretty unnecessary.

As for modifying Chart to serve the needs of mapbox data objects, I think it would be easiest and perhaps more maintainable to keep the methods separate but have the mapbox data-getting method append its results to the same app.dataCollection object. Does that make sense to you?

ptgott:
@ostermanj I won't be at the meeting tonight, but just want to say that having Chart constructors check for relevant data objects before loading them is a good idea. I'll check on what you/others get done tonight and help pick up any remainders later this week. I'll also work on constructors that are separate from Chart, fetch data for Mapbox maps and store it in app.dataCollection (checking to see if it's there first).

NealHumphrey:
@ostermanj and @ptgott are you guys wanting to merge this in yet, or still in discussion? Just want to make sure PRs get in when they're ready (but of course not before : )

ostermanj:
I'm happy to merge it now as the beginning scaffolding for getting the map onto the building page. We'll need to fix or replace the now-missing project.csv problem to go further with it, but that's true for all pieces.

@ptgott I have some local commits that have chart constructors automatically using an existing data object if it exists and creating it if it doesn't. Currently the developer has to know whether the data has been fetched already or not—not very sustainable. I tried out the sessionStorage but found two problems. One, it's slow: local storage has to be stringified on save and parsed on load; that and having to access the disk actually made it slower than fetching the data again. And two: I was having difficulty with the chart constructors doing their thing before the object from storage was ready. I couldn't figure that out because JSON.parse should be synchronous and have subsequent actions wait, but it wasn't working. Anyway, it might not be worth the performance hit, but maybe in production fetching the data would take longer than it is now (?).

PR #103, opened 2017-02-23
New Map files

rmcarder:
Added 4 files that are the start of a map-based dashboard. No changes to other files. Pie script works off the housing-insights.js constructor using the same format as moving-blocks.js. Ward buttons don't work yet, but will filter pie charts to show %'s of "YES" for each variable just for buildings in that ward, plus center the map on that ward.


NealHumphrey:
I'll let someone else decide when to merge this, @ostermanj or @ptgott ? I'm not up to speed enough on the javascript code structure

ostermanj:
@rmcarder this looks really good. A few things I noticed:

1. we lost the raw CSV file that we were linking to as DATA_FILE. @NealHumphrey we were linking to a project.csv as sort of dummy dat set to use.  Without that file, unfortunately, all of our D3 prototypes are broken—maybe you thought we were ahead of where we are on the data ingestion! oops.

If there's processed data ready to be ingested on the front end, we'll need learn how to do that (we haven't been contacting postgres or anything yet). Or if not can we just upload project.csv somewhere we can access it again?

2. ( @rmcarder ) it looks like the text block is grabbing data you've supplied to or through mapbox, with the map.queryRenderedFeatures method, but that the pie charts are trying to use project.csv, which is the same data. Would the mapbox data need to be updated manually if the project data were to change, or is it connected upstream somehow? It might be better to find a way to populate the text block by accessing the same data object the charts are using. If we could do it via D3 methods, even better.

3. I'd suggest putting the mapBox scripts currently in Map View.html into a separate .js file.

It would be nice to merge this as a an example of how to integrate our data with mapBox once the issues above, especial 1, are resolved. But overall we should be working toward having an integrated building.html page with all the elements present in the building page prototype. I'm afraid I cannot make the work session on Tuesday, but if enough of the team is there perhaps hashing out how to do that would be a good goal.

NealHumphrey:
@ostermanj and others - the /dev branch lost the projects.csv because over on the data team we abandoned the small_data (stuff to sync to Github) / big_data (stuff too big for Github) split in favor of only syncing data via S3. Deciding when to put stuff in one or the other was an extra layer of confusion. 

The projects.csv file is available for download here: https://s3.amazonaws.com/housinginsights/raw/preservation_catalog/20160401/Project.csv

To access it locally, you should store it in `{repofolder}/data/raw/preservation_catalog/20160401/project.csv`

instead of wherever it was previously. The whole /data folder is gitignored. If you need any other raw CSV files, they're all available on S3. In the near future (2 weeks?) I should have a Docker file set up that includes all the sync tools, but in the meantime let me or @salomoneb know if you need access to other raw data.

NealHumphrey:
Didn't realize you all were using the projects.csv directly, sorry to break it w/ the Python team's merge!

rmcarder:
Thanks John. To address #2, yes, so far I have just loaded the data into into Mapbox Studio and its being loaded again separately as a csv. It does not (and should not) happen this way; rather it was just a way to easily get data on there for now to start building out the idea. Data layers can be added to the map in the javascript from the data we are loading for the rest of the project. In this scenario we would just be getting a blank basemap from Mapbox so nothing would need to be separately updated there. I think actually, seeing today's comments, Paul may have already done this using the better method. I'll take a look at his latest commit tonight. Also yes, agree with 3, and with focusing on the building.html view. The reason I put it in a separate file is so that my pull request was only new files which I thought would be easier to merge than looking through the edits to the existing files.

PR #109, opened 2017-02-27
In the building page, in the sample Mapbox map, add a layer of dots f…

ptgott:
…or all public housing and another layer for the development in question.

Position elements of a panel within the building page using Bootstrap

Make parts of MapboxPortal (building-map.js) more extensible

Make it possible to remove the target building from a Mapbox source when layering sources on the building page

### DIFF
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -1,0 +1,44 @@
+GEM
+  specs:
+    addressable (2.4.0)
+    colorator (1.1.0)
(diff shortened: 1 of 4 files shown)


ostermanj:
@ptgott I'm up in Vermont with only my phone but from what I can tell this all makes sense. Let's merge it so it's available ahead of Tuesday. I can get back up to speed when I get back in town , Wednesday.

PR #115, opened 2017-03-01
Added initial subsidy timeline chart

wassona:
This needs some commenting and some styling, but is mostly complete and uses up to date prototypical inheritance. It also will need a decent bit of rewriting once we get our final data set. It references a data file that I shared in the D3 Slack channel but that I didn't commit to git.


wassona:
I keep thinking I have github down and being proven wrong. I thought I was working from a totally up to date branch, but there are some old, unnecessary changes to prototype-1a that we don't need in there. Also, apparently I'm using the wrong version of bundler. I should throw the Gemfile.Lock in my global .gitignore.

wassona:
I left a slight change in prototype-1a that makes the slider work better with the Metric selections. I think the merge is ready to go.

NealHumphrey:
@ostermanj it looks like the PR now contains just the files you mentioned, plus the additional prototype-1a change @wassona mentioned. What do you think, good to merge?

ostermanj:
Yes, I think so! Sorry for the delay. Been unusually swamped lately.

PR #116, opened 2017-03-02
Simplifies parameters in constructor call; auto checks for existing data object

ostermanj:
This request removes the data_name parameter from the constructor calls and does not require coders to know whether the data object a chart needs already exists or not. The Chart constructor in housing-insights.js will check for an object keyed by a hashed value of the DATA_FILE requested. If it exist, the object is used; if not, new data is fetched and a new object made. In practice, the constructors are called and made so  quickly that object is recreated several times. This will be more useful for when a chart is called after some sort of user action, like a click or hover.


ostermanj:
@ptgott @NealHumphrey @wassona @rmcarder  I've updated this PR to have a rough but working way to get a specific building's data from a query string in the url. For example, clicking on a square on /moving-blocks.html will navigate to /building.html with the represented building address id in the query string. rough code in building-header.js grabs that query string and uses it to populate the fields in the header.

It is not connect in any way to the map view, and some of the data in the header (phone number, tax assessment) are not in the DATA_FILE.

Sorry to miss the session tonight. See you all soon.

PR #118, opened 2017-03-06
Handle multiple data sources in the building information page

ptgott:
@NealHumphrey @vincecampanale @wassona @ostermanj 
I've added a grabData() function that requests geoJSON from data sources specified in a 'datasets' object literal. A setInterval() function waits for all responses to come back, then assigns the geoJSON to the 'data' keys in the 'datasets' object, and calls a prepareMaps() function that produces the Mapbox maps.

Doing this created issues with scope in defining layers in the MapboxPortal constructor, so I've moved all layer definitions to prepareMaps(). Now MapboxPortal simply iterates over all the layer IDs passed to an argument as an array of strings.

I've also added a 'buildingForPage' object literal as a placeholder for whatever way we decide on to bind data about a selected building to the information page.

At the moment, there is an error in loading the maps. As far as I understand it, this is because Mapbox maps can't include a source with the same ID multiple times.

I wanted to issue a pull request to keep you all up to speed on what I'm doing, but I can work on fixing the error before we merge if that's what you'd prefer.

### DIFF
--- a/javascript/tool/building.html
+++ b/javascript/tool/building.html
@@ -83,13 +83,27 @@
           </div>
           <div class = "row">
             <div class = "col-sm-6">
-              <div class="mapbox-portal" id="test-map"></div>
+              <div class="mapbox-portal" id="affordable-housing-map"></div>
             </div>
             <div class = "col-sm-6">
               Here's some test text.
             </div>
           </div>
         </div>
+        <div class="container">
+          <div class = "row">
+            <h3 class = "col-sm-12">Metro stations nearby</h2>
+          </div>
+          <div class = "row">
+            <div class = "col-sm-6">
+              <div class="mapbox-portal" id="metro-stations-map"></div>
+            </div>
+            <div class = "col-sm-6">
+              Here's some test text.
+            </div>
+          </div>
+        </div>
+
       </section>
     </section>
 
--- a/javascript/tool/js/building-map.js
+++ b/javascript/tool/js/building-map.js
@@ -70,13 +70,38 @@
     }
   }
   
+  // The MapboxPortal constructor inserts a Mapbox map into element with elementID.
+  // It always adds certain layers (the dot and label for the building-of-interest).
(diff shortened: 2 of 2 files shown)


ptgott:
Also, I forgot to squash. Sorry about that!

ostermanj:
@ptgott the mapbox stuff would take me a while to get up to speed on, but I do see what you're doing here. No luck with a success callback finction instead of the setInterval check on the data collection? 

I'll approve so you can merge tonight. Sorry I won't be there.  As for how to bind to a specific building's information, could be done by aquery parameter in the url. I have a rough example work in [PR 116 ](https://github.com/codefordc/housing-insights/pull/116)

ptgott:
@ostermanj, @wassona, @vincecampanale 
I've pushed another commit to this pull request. The commit removes the problem with multiple Mapbox sources conflicting and throwing 'source not found 'errors. What I've done is include a ProtoLayer object that bundles together two object literals that follow Mapbox conventions: one for  a layer and one for a source. The reason I've done this is because Mapbox layers specify their sources with an 'id' only. You have to refer to a source in addSource(), which I've placed in the MapboxPortal constructor. However, since MapboxPortal is outside the scope of prepareMaps(), it doesn't have access to the 'sources' object, which depends on an initial round of XHR requests. Having MapboxPortal take ProtoLayers as an object seems to solve the scope problem. I suppose having 'sources' be scoped to the IIFE, rather than prepareMaps(), would also solve this problem. If that makes more sense I could tweak it.

Oh, and to answer @ostermanj's earlier question about using a success callback vs setInterval(): I had originally created each XMLHttpRequest object as the value of a key in a 'requests' objects literal. Each request would have a 'readyState == 4' callback that would iterate through the other 'requests' object and check whether the other requests objects also had the same readyState. This resulted in some strange behavior (only one requests object would be recognized as loaded), so I went with the setInterval() approach.

ostermanj:
Thanks, @ptgott . Good explanation.

PR #121, opened 2017-03-08
Added accordion functionality to building display page

wassona:


ostermanj:
I think I resolved the merge conflicts correctly. Still can't wrap my head around the notations very well, but  the accordions do  appear to working as intended.

ostermanj:
@wassona fyi your data file for the timeline page looks like it's missing from the codebase: './project_sample_subsidy.csv'

Issue #80, opened 2017-02-12
Adding an information page for individual buildings


ptgott:
@wassona @vincecampanale As we talked about on Tuesday, here's a separate Issue for coordinating our work on the building page. We can refer to this when making commits related to the building page.

One thing to think about is how we will bind the data relating to a given building, as loaded in the moving block diagram, to that building's page. We don't have to determine this until after we've produced the visualizations for the accordion. In the meantime, we could think about using a 'dummy_building.json' object to refer to, or integrating such an object into a .js file.

vincecampanale:
I think 'dummy_building.json' is a solid solution for now. We could even get real data from our data set and populate `dummy_building` without that data.

NealHumphrey:
Let us (data team) know what format works best for you for the data, and we can make sure our python scripts output it in that format! We can make a couple real buildings worth for you to work from, too.

ptgott:
@hansak11 @vincecampanale @NealHumphrey @wassona @ostermanj @jeannejk
Here's a list of data fields, as we talked about. They all come from the prototype of the building information page. We can add to/edit this list as we work through the building page, find data sources etc!

### Coordinates of non-building things; used for determining what's near a building
Metro stops
Bus stops
PUD applications
Building permits
Buildings with affordable housing
### Information to include for a given building's page
Subsidy types (Sec 202, Sec 8, Sec 207/233) and when they expire
Non-profit owned
Number of units
Number of studios
Number of 1BRs
Years until next subsidy expires
REAC score
Address
Owner telephone
Tax assessment

Jeannejk:
This is a first attempt at finding these datafields in the tables loaded into postgres. I found the source for some fields unclear, which I included in a notes column. For now I've been collecting the info in excel, but we can move it into a better format at the next meeting if needed.

[Data Field Sources.xlsx](https://github.com/codefordc/housing-insights/files/789048/Data.Field.Sources.xlsx)

ptgott:
#109 
I've done a bit more work on the Mapbox part of the building information page. So far I've pulled data from Open Data DC (the 'Affordable Housing' dataset) for development purposes, in place of the old Projects.csv. I've added a layer in one Mapbox map for all affordable housing. I've also done some work on making the MapboxPortal constructor a bit more extensible. 
### Here are some things that are left to do in adding Mapbox maps to the building page:
1. Integrate the building page with the 'app.dataCollection' object
[...]

PR #135, opened 2017-03-18
On the building page, list nearby features

ptgott:
#80 
@ostermanj @wassona @vincecampanale 

Since the text of the map-related accordion panels in the building page prototype has to do with features near the target building, I've copied some code from a library for geospatial calculations to assign distances to a given Mapbox source and list those distances next to the relevant map.

A big difference from the prototype is that this code calculates geospatial distance to nearby metro stations, whereas the prototype shows walking distance. While the calculations of geospatial distance use snippets of a JS library copied to a file in the 'js' folder, I imagine calculating walking distance will involve some asynchronous API calls to, say, Google Maps. This can be something to look into later!

### DIFF
--- a/javascript/tool/building.html
+++ b/javascript/tool/building.html
@@ -104,8 +104,7 @@
 						<div class="col-sm-6">
 							<div class="mapbox-portal" id="affordable-housing-map"></div>
 						</div>
-						<div class="col-sm-6">
-							Here's some test text.
+						<div class="col-sm-6" id = "nearbyAffordableHousing">
 						</div>
 					</div>
 				</div>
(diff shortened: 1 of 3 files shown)


NealHumphrey:
@ptgott you should know we have some Python code that's already set up to do those walking distance calculations. It's not totally integrated into our new pipeline yet so it's not technically in the database yet, but it's close. Here's the code: https://github.com/codefordc/housing-insights/pull/87

There may still be some use for this client-side distance calculation. But, we should discuss where this best fits. For discussion next meeting!

NealHumphrey:
Hey @ptgott and @ostermanj  - what do you think about the best thing to do with this pull request? I think we should definitely keep this somewhere since we may end up wanting to do some client-side distance calcs, though for the specific use we should use the Python version which will eventually be available through the API Priscilla is working on. Merge in as is? Or move into an examples folder somewhere? Leave as an  independent branch?

ptgott:
What would you say to merging it in with the assumption that eventually we'll replace it with API calls (i.e. and delete it)? This way it'll be available if we need to use it (through a previous commit in the log), but it won't take any space in the project directory if we don't.

PR #151, opened 2017-03-24
Fixes s3 automatic download when user doesn't have local data

NealHumphrey:
- Fixes bug for downloading the file from S3 when it does not exist locally.
- Adds ability to load the file directly from S3 without a local download. This should only be used when running on AWS directly (in the future)


NealHumphrey:
@louvis Got it fixed!

Issue #143, opened 2017-03-22
MVP: Integrating elements into a map-based minimum viable product


ptgott:
At the 21 March meeting we talked about producing a basic minimum viable product that we could easily extend with various data sets. Some of the work consists in stitching existing components together, and some requires adding some new features. Some requires work on the backend. 

We'll organize the stitching around the Map View (javascript/tool/map-view.html). The user will select from geographic aggregation types (neighborhood, ward, zip code...) and datasets (rent, crime...) to display. There will be a menu for aggregation type/region; a menu for datasets; a map (the existing Mapbox map in map-view.html); and a sidebar for aggregations (the existing sidebar in map-view.html).

There are five main tasks:
- [ ] Turning the building information page into a partial that hovers over the map view. Clicking on the hover panel brings the user to the building page. 
- [ ] A menu for aggregation/zone types that (a) change the data that display in the map and (b) change the aggregation charts in the side panel of the Map View.
- [ ] A menu for datasets (e.g. crime and rent) that (a) produces a heatmap of aggregate data across the Map View and (b) produces donut (and other?) charts of aggregates on a side view.
- [ ] Ensuring that names for fields are consistent across our queries and checking them against meta.json. (copy meta.json from the data folder to the javascript folder). Whenever a particular data field is displayed in a graphic, pull the 'display_name' and 'display_text' from the meta.json rather than hard-coding the data name directly. 
- [ ] (Neal) Getting data transposed into different formats, aggregation/zone types (e.g. neighborhood cluster level, zip code level)

Let me know if this all makes sense and feel free to adjust the wording of the features!

NealHumphrey:
I added a slight clarification to the meta.json  bullet. 

I also put my name by the one that I'll work on (transposing zone-based data to different zones). I reached out to Peter Tatian on this as he does something similar in Neighborhood Info DC; I also think I have a good idea on how to do it using DC's Master Address Repository (i.e. using residential housing units as a proxy for population, since we have residential housing units at the exact-address granularity level so can wrap them up into any arbitrary zone type).

ptgott:
Unless anyone else is already working on it, I'll start on integrating the building information page into the new layout (the first task).

ostermanj:
Yeah, go for it!

ptgott:
I've done some work on linking the Building Page to the Map View in #153 
### What I've done
1. I've done a bunch of minor whitespace changes. This is why the pull request might seem more extensive than it really is.
[...]

PR #160, opened 2017-04-02
Add a method of 'app' that fetches tables from the AWS API using AJAX…

ptgott:
See comment in issue #143

### DIFF
--- a/javascript/tool/js/housing-insights.js
+++ b/javascript/tool/js/housing-insights.js
@@ -55,19 +55,21 @@ function APIDataObj(json){
   
   // this.geoJSON assumes that we'll be producing a FeatureCollection with Features
   // that are points. For shapes, we may want to add an argument and code that responds to it.
-  // This code also assumes that each element in 'json' has a 'longitude' and 'latitude'
-  // property.
-  this.geoJSON = function(){
+  // Currently this method requires us to specify the latitude and longitude fields
+  // of the data within the json. If we standardize the json to always call its geospatial
+  // fields 'longitude' and 'latitude', this may not be necessary.
+  this.geoJSON = function(longitudeField, latitudeField){
     var features = json.map(function(element){
       return {
         'type': 'Feature',
         'geometry': {
           'type': 'Point',
-          'coordinates': [+element.longitude, +element.latitude]
+          'coordinates': [+element[longitudeField], +element[latitudeField]]
         },
         'properties': element        
       }
     });
+    console.log('json', json);
     return {
       'type': 'FeatureCollection',
       'features': features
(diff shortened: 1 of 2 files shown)


NealHumphrey:
@ostermanj are you able to review this one? Now that we have more Python contributors I'm hoping to get fully up to speed on Javascript/D3 code next week so I can do better PR reviews.

javascript/tool/js/housing-insights.js
```diff
@@ -46,9 +46,97 @@ String.prototype.hashCode = function() {
...
+    }
+  }
+  
+}
+
 
 var app = {
     dataCollection: {}, // empty object to house potentially shared data called by specific charts, see above
+    
+    // getAPIData exists to fetch data from the AWS API based on meta.json. This is a separate
```
NealHumphrey: I think this sounds like a great approach - keep both methods available side by side as we transition. I don't see a problem with keeping it named getAPIdata even after getData is deprecated. 

I tested this locally and worked. I am going to modify this to instead use the Flask API demo after merging.

javascript/tool/js/housing-insights.js
```diff
@@ -46,9 +46,97 @@ String.prototype.hashCode = function() {
...
+    // can change it as the Python side of the project develops.
+    // This function adds the resulting json to dataCollection.
+    
+    // 'tableNamesArray' is an array of strings, each element corresponding to the name
+    // of a table in meta.json in the '/python/scripts/' directory.
+    // This will allow us to keep all of our asynchronous requests in one place so that 
+    // we can use them for maps and charts when they finish.
+    
+    // 'doAfter' is a callback where we can specify all the specific constructors to call with the data
+    getAPIData: function(tableNamesArray, doAfter){
```
NealHumphrey: We will want of course to think about how to expand this method, or to add additional similar methods, as we expand the types of data available in API endpoints - not all our endpoints will follow the 'get all records in this tablename' format.

NealHumphrey: For discussion next meeting perhaps

javascript/tool/js/rich-map.js
```diff
@@ -9,187 +9,197 @@ var map = new mapboxgl.Map({
...
+        'circle-radius': {
+          'base': 1.75,
+          'stops': [[12, 3], [22, 180]]
+        },
+        // color circles by ethnicity, using data-driven styles
+        'circle-color': {
+          property: 'category_code',
+          type: 'categorical',
+          stops: [
+            ['1 - At-Risk or Flagged for Follow-up', '#f03b20'],
```
NealHumphrey: After Monday's demo when we add other point-based sources, we'll want to find a way to transition this to be a generic categorical color map that can accomodate any other point data source, or to store these type of settings on a per-data set basis so that things get consistently represented across multiple pages of the app. Just flagging for later discussion.

NealHumphrey:
Merged via the command line to deal with merge conflicts. 

@ptgott and @ostermanj I'm pretty sure I resolved the merge conflicts appropriately, but do take a look at the current code and make sure it does what both of you expect.

Issue #149, opened 2017-03-23
POC: Link up to the Master Address Repository


NealHumphrey:
We need to connect to the Master Address Repository for various reasons:
- Proxy for population-weighted zone allocations (count all residential units in each zone type, e.g. Ward or Zillow Neighborhood)
- Finding the unit counts of specific buildings. 
- Geocode addresses into zone types when one is not provided

Set up a proof of concept of how to query the MAR for one/all of these scenarios:
- Provide an address, return the unit count. Demo using [project.csv](https://s3.amazonaws.com/housinginsights/raw/preservation_catalog/20170315/Project.csv)
- Use the 'block' field from the 'crime' table to find out the zip code for each crime entry (start with a small sample of 10 crimes). 

Bonus example (lower priority):
- Count all residential addresses in a specific zone (e.g. zip code 20009). Be sure to limit the results of this during development to the first couple of results to avoid overloading their servers. Eventually we will want to be able to query all resulting residential addresses in the city so that we can use non-DC zone types (e.g. Zillow neighborhoods). 

Documentation:
https://docs.google.com/document/d/1BJSeQ7Ghd6TPd22tVsYr-ATmsh3Eve0rJMrlcU2mAis/pub

NealHumphrey:
Resolved in #164!

Issue #49, opened 2016-12-28
Prototype #2: Dashboard Library


NealHumphrey:
After completing #48 (Prototype number 1), we'll rearrange some of the elements and make some new ones to make a second prototype, focusing on the 'Library' view that includes a dashboard quick-view of each of the library analyses. 

This one will likely require a bit more playing with our actual available data to come up with some useful analyses, as we'll be testing the home page method (i.e. the library) as much as we're testing the individual graphs contained therein. We will make this prototype second; while the prototype team is working on Prototype 1, some people can start working with the data to generate options for the sample analyses we should include in the library.

NealHumphrey:
Closing this issue - we'll make a new one based on our interviews

PR #164, opened 2017-04-04
W.I.P. -- Master Address Repository #49

ehenry2172:
Simple implementation of #149. It currently provides a script to go retrieve a csv for a given address. This is not finished, but P.R. submitted to receive feedback from the Tuesday meeting. Its an attempt at making a modular interface so other data sources can rapidly be added and data retrieved.

Example Usage:
bin/data.sh mar find_location --output ~/output.csv --params "location:Some street Address"


NealHumphrey:
One other edit. Just to avoid confusion between *our* API and getting data from *other* api's, can you either move object based code into `/ingestion/api` or rename the `/api` folder to something like `/api_requests` or `/api_data` (both those names are clunky - I'm open to better suggestions). We will soon be adding code to serve our data through an API server that we set up so I want to keep the distinction clear.

ehenry2172:
I definitely hear you in regards to the opacity of the code -- to do it in a modular way and quickly I used a TON of reflection. I'll create a README to drop in the directory to explain the layout and how to "Create your own api class", as well as including an example on how to call the Api Connection functions directly, similary to what data.py does.

Just to clarify, my two end goals are to be able to get output csv's by using **data.sh** - - data.py should never be used directly - - and to have a Api Connection class for each api that other scripts can import if they want to have direct access to that api.

I'd like to move the **/api** directory to **/sources** , to imply that these are ways to connect to data sources. I want to keep it separate from the /ingestion module because my approach is significantly different.

NealHumphrey:
`/sources` works for me!

I see how the shell file connects now - I'm not very proficient  with using these to run python myself. Mostly for clarity, I still recommend putting an example that has a hard-coded example address (you could use the 641 S St NW address of where we're meeting tonight) and hard-coded names of the api module and method. The data.sh can still be the generalizable version available in addition to a hard coded example . 

Thanks!

ehenry2172:
Thanks for the additional feedback and testing - - that's very helpful. I'll work on remediating these issues tonight.

ehenry2172:
I've added a bunch of documentation and fixed most of the items you mentioned with the exception of dealing with not found addresses and get_condo_count. I would recommend that you go ahead and merge it and i'll submit another pull request next week with some updates.

NealHumphrey:
@ehenry2172 I added a new issue: https://github.com/codefordc/housing-insights/issues/175. I added a little bit of additional documentation on how to run the data.py directly (since I'm on Windows, and a few others are as well). Be sure to pull the latest dev when you do so. 

Can you add in a few unit tests as well when you're fixing those issues?

Thanks, this is great!

PR #174, opened 2017-04-05
Add method to create folder if it doesn't exist

danshorstein:
I wasn't able to test this as I haven't fully figured out how yet, and it doesn't look like a unit test is setup yet. But this could work, so try it out and let me know if it's close but needs some tweaks.

### DIFF
--- a/python/housinginsights/ingestion/DataReader.py
+++ b/python/housinginsights/ingestion/DataReader.py
@@ -172,6 +172,16 @@ class DataReader(HIReader):
 
         super().__init__(self.path, self.path_type)
 
+    def validate_or_create_path(self):
+        root_path = os.path.abspath(os.path.dirname(self.path))
+        sub_root_path = os.path.abspath(os.path.join(root_path, '..'))
+        if os.path.exists(root_path):
+            return
+        if not os.path.exists(sub_root_path):
+            os.mkdir(sub_root_path)
+        os.mkdir(root_path)
+        return
+
     def download_data_file(self):
         '''
         Checks to see if the file already exists locally, if not downloads it
@@ -181,6 +191,7 @@ class DataReader(HIReader):
                 myreader=csv.reader(f,delimiter=',')
                 headers = next(myreader)
         except FileNotFoundError as e:
+            self.validate_or_create_path()
             logging.info("  file not found. attempting to download file to disk: " + self.s3_path)
             urlretrieve(self.s3_path, self.path)
             logging.info("  download complete.")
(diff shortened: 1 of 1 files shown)


danshorstein:
FYI after looking at the slack conversation, this method probably needs to be tweaked to include creating any sub folders. I think it could iterate through the sub folders of the path, check to see if each one exists, and create it if it doesn't exist.

NealHumphrey:
@danshorstein I just verified this works for a single level of folder nesting, but does not work with nested folders. 

If you want to test it locally, you can [follow the instructions](http://housinginsights.org/resources/onboarding/docker.html) to rebuild the database. Then, delete a folder in the /data/raw folder that contains one of the data files listed in the manifest.csv. The code should download the file if it doesn't find it locally, so when you rerun the load_data.py function it should notice the folder you deleted and try to download it from S3. Currently, anyone that hasn't run the awscli sync command will be missing the crime data folder as that folder wasn't skeletoned out in the repo with .gitinclude files (that's why the other folders were already available - they came as empty folders as part of the repo). 

If you want to make a unittest, the tricky part is that we don't have the tests/test_data set up to have its data in S3. Probably easiest to do manual testing for now and in the near term I'll set up S3 to have a housinginsights_test folder for this type of thing.

danshorstein:
Ok I just figured it out - os.makedirs creates nested folders, woot! Creating a pull request momentarily.

NealHumphrey:
Also FYI, you can keep the pull request open and continuing adding commits to it. If you commit to the same branch and push it to your fork, this pull request will be updated with the additional commits.

NealHumphrey:
Also, you can manage your fork of the repo however you want, but I do find it useful to always make a new feature branch rather than committing directly to your dev branch. That way you can keep your dev branch the same as our dev branch in case your pull requests don't get merged in right away (if they have issues etc.) Just a suggestion

danshorstein:
Awesome, just did that and it worked, thanks. I'm very new to using github, so thanks for bearing with me =)

danshorstein:
And I'll try creating the local database and testing later. I probably need to use my macbook, as my windows PC does not seem to want me to use docker.

NealHumphrey:
Hmm, yeah - which Windows are you on? I was able to get my Windows machine working with it, but when I tried to set it up on my slightly older laptop (to test the onboarding process) it didn't work there. 

If you want to set up your Windows laptop, instead of using Docker you can also just install Postgres directly - Postgres 9.5 is all you need to do the rebuild of the database. In `secrets.json` there is an option for "local_database" instead of "docker_database" - you'll just need to create a database in your Postgres installation and edit that connect string to match your Postgres port number, username/password, and database name. 

Docker's useful in that once you get it running, you can usually run most any docker image out there, but I've found it to be not quite as plug and play as people portray it to be especially on Windows. Once it's running though, it makes life a lot easier especially as the project evolves and we add additional dependencies.

NealHumphrey:
@danshorstein I just verified this now works for nested folders. Thanks! Merging.

Issue #130, opened 2017-03-15
Home Mortgage Disclosure Act - research data


NealHumphrey:
Research what data is available via the Home Mortgage Disclosure Act. This has been recommended by a couple of people that we have interviewed or that have heard about the project. 

- How granular can we get data (city level, census tract level, etc.)
- What data fields are available? What caveats are there to them? What do they tell us about the city?
- What time periods is the data available for?
- What data format is it provided in? 
- Where are updates to the data posted, and how frequently?
- What fields are available to connect this data to our other data sets? e.g. zip code, census tract, etc. Do we need to create any sort of cross-walk table to translate between them?

As soon as the data ingestion pipeline is set up, we can also download and integrate them into the current system (can make a separate issue for that when the time comes).

NealHumphrey:
@emkap01 This might be a good next task for you to look into while we wait to decide what to do about the PUDs! Let me know if you want to grab it (you can assign yourself using assignees if so)

emkap01:
I grabbed it. I'll start looking into it.

emkap01:
I've started a [draft report](https://docs.google.com/document/d/1xcPpX37nq4TUpb-YNcDM0U915z0Eq7DkALohKOPJeG8/edit?usp=sharing) which includes a link to a sample of the data.  I'll continue updating it as I learn more. Based on what I've seen so far, it looks like the HMDA records could be useful for establishing mortgage trends at the census tract level.  There were ca. 35,000 HMDA records for DC in 2015.  It's available via a well-documented API and ostensibly includes records back to 2007.

NealHumphrey:
@emkap01 cool, good stuff. I watched the video. So it looks like each record is an individual loan application (accepted or denied)? Did you get a sense of whether the data is complete (i.e. all loans in the US) or if there's any particular bias to which types of banks submit to HMDA and which types of loans they submit?

emkap01:
Yes, each record represents an individual loan application, regardless of whether it was accepted or denied.  The full dataset contains ca. 35 million records and ostensibly represents the entire U.S.  There is a field labeled 'purchaser_type_name' that indicates which type of entity bought the loan, however it only contains a single value to describe any type of bank - ' Commercial bank, savings bank or savings association' - so it looks like it's not necessarily possible to sub-categorize banks in this dataset.

[...]

PR #181, opened 2017-04-07
#130 - HMDA upload

emkap01:
I think/hope I was able to roll the marble end-to-end and add 250MB of Home Mortgage Disclosure Act records to the project, per issue #130.  

I followed all steps in the [how-to guide](http://housinginsights.org/resources/adding-data-sources.html) to get the records cleaned and added to the Postgres db, as well as push a copy of the raw records to the S3 bucket.   I did *_not_* attempt to join the new table to any other tables in the db, although I _did_ add a function to my cleaner class to make sure that the value we can use for that join (census_tract_number) is consistent with what's already in the db.  I will play around with the joins next.

This is the first pull request I've submitted on this project, and I'm very much an amateur at Python, so please don't hesitate to let me know if I messed something up.


docker-compose.yml
```diff
@@ -10,14 +10,7 @@ services:
             - ./docs:/srv/jekyll
         ports:
             - "4000:4000"
-    web:
```
NealHumphrey: @emkap01 can you remove this file from the pull request? We modified this version to get it to work locally for you, but don't want that in everyone's version. Here's instructions: http://stackoverflow.com/questions/12481639/remove-files-from-git-commit

If you're not comfortable doing this, we can do it in person on Tuesday instead.

python/scripts/meta.json
```diff
@@ -2083,6 +2083,558 @@
...
+                "source_name": "application_date_indicator",
+                "sql_name": "application_date_indicator",
+                "type": "text"
+            },
+            {
+                "display_name": "as_of_year",
+                "display_text": "",
+                "source_name": "as_of_year",
+                "sql_name": "as_of_year",
+                "type": "text"
```
NealHumphrey: should this be a date? Or an integer?

python/scripts/meta.json
```diff
@@ -2083,6 +2083,558 @@
...
+                "source_name": "loan_amount_000s",
+                "sql_name": "loan_amount_000s",
+                "type": "integer"
+            },
+            {
+                "display_name": "number_of_1_to_4_family_units",
+                "display_text": "",
+                "source_name": "number_of_1_to_4_family_units",
+                "sql_name": "number_of_1_to_4_family_units",
+                "type": "decimal"
```
NealHumphrey: should this be integer?

python/housinginsights/ingestion/Cleaners.py
```diff
@@ -308,6 +308,21 @@ def replace_tracts(self,row,row_num,column_name='census_tract'):
             logging.warning('  no matching Tract found for row {}'.format(row_num,row))
         return row
 
+    def append_tract_label(self,row,row_num,column_name='census_tract_number'):
```
NealHumphrey: This looks great!

NealHumphrey:
@emkap01 I was able to get this to run on my machine and successfully imported the data - good stuff! Just a couple minor edits and this will be good to go.

docker-compose.yml
emkap01: I took a shot at updating the pull request in order to remove docker-compose.yml via the instructions in that link, let me know if it didn't work.

python/scripts/meta.json
emkap01: The values in this field will consistently be a 4-digit year.  I wasn't sure if they should be dates or not, so I cast them as text.

python/scripts/meta.json
emkap01: No, I examined a subset of these values as a spreadsheet and some of them are actually decimals.

python/housinginsights/ingestion/Cleaners.py
emkap01: Thanks!  This was about as far as I could stretch my amateur Python skills, but I was really excited to discover that it worked when I looked at the records in my local copy of the db.

python/scripts/meta.json
NealHumphrey: Let's make this an integer then - that way we can do comparison (>, < ) etc. on the years if needed.

NealHumphrey:
@emkap01 it doesn't look like any new commits have been added - have you pushed your upates yet?

emkap01:
Apparently I screwed something up - it looks like I overwrote the files in my pull request when I was trying to remove the docker-compose.yml file, because the versions I included in the pull request no longer exist in my local or remote repos.  If you have the ability to resolve the issue from your end, go ahead, otherwise we can just fix it on Tuesday night.

NealHumphrey:
@emkap01 I can use the copy I pulled to fix it - will do that before Tuesday's session and we can make sure it's all squared away then.

terrysky18:
I've got the docker-compose.yml file on my repository and local drive.  Let me know if you need it.

emkap01:
As discussed I'm just going to close the pull request, delete the branch, then create a new branch with the correct files and submit a new pull request based on that.

PR #182, opened 2017-04-07
Fix docker-compose race condition Fixes #173

### LINKED ISSUE #173: Race condition in Docker file
The `docker-compose.yml` file calls three services - jekyll, web (a postgres web client) and postgres (the postgres database). Docker appears to start all 3 services concurrently. However, the web client needs postgres to already be running in order to load correctly. There is a race condition that means if Postgres happens to start first all is well, but if pgweb starts first it fails and the user has to restart the container with docker-compose. 

[This solution I found](http://deployeveryday.com/2016/09/12/race-condition-docker-compose.html) uses a Python script to solve the problem by having docker-compose call the Python script, and Python handling the waiting until Postgres starts. For us to use this approach, we would need to add Python to our docker compose - which we will need to do soon anyways to run our API web client (probably Flask). If you go this route, you can use this Python 3.5 script runner image:
https://github.com/codefordc/project-management/tree/master/python

Note, the example solution linked uses the Docker file directly instead of docker-compose, so solution of how to actually call the image in conjunction with the other images will be slightly different.

jasonrhaas:
Fixes the docker-compose race condition by re-trying in the event of a failure.  However, the retries are bounded (10 for web, 3 for postgres) so that if there is an actual problem with a configuration, it will still fail as expected, rather than retrying infinitely.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,8 @@ services:
             - postgres
         ports:
             - "8081:8081"
-      
+        restart: "on-failure:10"
+
     postgres:
         image: codefordc2/housing-insights-postgres
         environment:
@@ -27,3 +28,4 @@ services:
             - POSTGRES_DB=housinginsights_docker
         ports:
             - "5432:5432"
+        restart: "on-failure:3"


NealHumphrey:
@mkalish this should probably get added upstream to the project-management repo too.

NealHumphrey:
@mkalish I went ahead and did a pr in that repo

PR #184, opened 2017-04-07
Add make_draft_json unit tests

jasonrhaas:
## What's New

Added some units tests for the `make_draft_json.py` file.  There are still more tests that can be added to the `test_make_draft_json(self)` and `test_make_all_json(self)` methods, but this is a good start.

## Testing

Running the full ` nosetests --verbosity=2 --with-coverage --cover-inclusive --cover-erase tests` command doesn't work because there is an import error in the `test_ingestion.py` file.  This needs to be fixed.

However, the tests I added pass when running individually.

`nosetests -v tests/test_make_draft_json.py -v`

```
nose.config: INFO: Ignoring files matching ['^\\.', '^_', '^setup\\.py$']
TODO:  Still needs more tests. ... ok
TODO:  Still needs more tests. ... ok
test_pandas_to_sqal_data_type (python.tests.test_make_draft_json.TestMakeDraftJson) ... ok
test_sql_name_clean (python.tests.test_make_draft_json.TestMakeDraftJson) ... ok

----------------------------------------------------------------------
Ran 4 tests in 0.006s

OK
```

### DIFF
--- a/python/tests/test_make_draft_json.py
+++ b/python/tests/test_make_draft_json.py
@@ -1,0 +1,53 @@
+import unittest
+from unittest.mock import patch, MagicMock
+from housinginsights.ingestion.make_draft_json import (sql_name_clean, pandas_to_sql_data_type,
+                                                       make_draft_json, make_all_json)
+
+
+class dataframe_file:
+    columns = ['foo', 'bar']
+
+
+class TestMakeDraftJson(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


jasonrhaas:
@NealHumphrey We can definitely implement Travis CI, I'm a big fan.  It can probably be set it up fairly quickly.  And we can still use the test files in Travis, it will just pull from whatever is in your Git Repo.

PR #185, opened 2017-04-10
Issue 169 encoding param

speedyturkey:
This is my first PR so please be gentle - I believe I understand the DataReader classes and that this code will work. I did not think it necessary to implement a new __iter__ method on the DataReader class, but rather modified that method on HIReader.

Probably no need to keep the add_encoding_to_meta_json.py script but thought I'd share anyway.

I'm grateful for any feedback and happy to tweak further if necessary.

### DIFF
--- a/python/housinginsights/ingestion/DataReader.py
+++ b/python/housinginsights/ingestion/DataReader.py
@@ -26,18 +26,19 @@ class HIReader(object):
     File can be local (path_type="file") or remote (path_type="s3"). Note, local files are preferred
     when possible for faster processing time and lower bandwidth usage. 
     """
-    def __init__(self, path, path_type="file"):
+    def __init__(self, path, path_type="file", encoding="latin-1"):
         self.path = path
         self._length = None
         self._keys = None
         self.path_type = path_type
+        self.encoding = encoding
 
     def __iter__(self):
         self._length = 0
         self._counter = Counter()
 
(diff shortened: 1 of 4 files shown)


speedyturkey:
Added the requested changes to make_draft_json. Do I understand correctly that when running make_draft_json on the full manifest, that it generates individual json files? Should that function generate a complete, unified json instead?

NealHumphrey:
@speedyturkey heads up, when running this locally I realized there's some conflicting implentation of this. It's the confusing manifest vs. meta issue. Since manifest has rows on a per-file basis, we want to put the encoding definition there. Your DataReader class looks for the encoding param in the manifest_row class, but you actually put it in the table metadata. 

In the data reader, when you do manifest_row.get('encoding', 'latin-1'), the silent fallback default hides the fact that this param was never actually found. I realized this because I wanted to add logging anytime the fallback was used, since we want to avoid silent bugs when people don't correctly add a data source. 

I'll fix this in the current dev since I've already merged this PR. Just wanted to give you a heads up if you look at this code again later.

NealHumphrey:
@speedyturkey fyi, here's the commit that resolves this: https://github.com/codefordc/housing-insights/commit/0006f5d456657012551242e1304b2ab14f846362

speedyturkey:
Great catch, thanks for pointing out. I like your solution.

PR #186, opened 2017-04-10
Added REAC Scores data from Preservation Catalog

louvis:
@NealHumphrey see if I managed to submit this correctly. I think I ended up adding some changes for .gitkeep files related to data syncing that shouldn't be merged.


ehenry2172:
You got caught by the invisible Micro$0ft carriage return. Try running this from your command prompt, it should instruct git automatically remove those if you miss them by accident: git config --global diff.astextplain.textconv "astextplain"

NealHumphrey:
@ehenry2172 good catch. But what about the `autocrlf` setting instead? I hadn't seen the `diff.astextplain` 
 setting before, but it looks like that does more than just line endings. Here's the autocrlf setting:
`git config --global core.autocrlf true` (and [where to read more](https://help.github.com/articles/dealing-with-line-endings/)

@louvis, I might try the autocrlf setting first and if that doesn't work then do the astextplain one. After you change the setting, you'll want to make a new commit and push it up - if that commit changes all those .gitkeep files back to the original line endings, you're good to go. If you push new commits to the same branch, this PR will be updated. 

And if you're curious about [what's going on](http://stackoverflow.com/questions/1552749/difference-between-cr-lf-lf-and-cr-line-break-types)

ehenry2172:
@NealHumphrey 's answer is probably what you should use as it's better documented. The setting I suggested is added by Git BASH when you tell it you want to use only Unix Line Endings.

python/scripts/manifest_sample.csv
```diff
@@ -1,4 +1,5 @@
...
-use,acs_rent_median_temp,acs_rent_median_09_5YR,7/1/2007,../../../data,https://s3.amazonaws.com/housinginsights/,raw/acs/B25058_median_rent_by_tract/2009_5year/ACS_09_5YR_B25058_with_ann.csv,Ready to go data at the census tract level. Do pay attention to margins of error
+include_flag,destination_table,unique_data_id,data_date,local_folder,s3_folder,filepath,notes
+use,project_temp,project_sample,4/1/16,../../../data,https://s3.amazonaws.com/housinginsights/,sample/project_sample.csv,Sample data
+use,acs_rent_median_temp,acs_sample,7/1/12,../../../data,https://s3.amazonaws.com/housinginsights/,sample/acs_median_rent_sample.csv,Sample data
+use,acs_rent_median_temp,acs_rent_median_09_5YR,7/1/07,../../../data,https://s3.amazonaws.com/housinginsights/,raw/acs/B25058_median_rent_by_tract/2009_5year/ACS_09_5YR_B25058_with_ann.csv,Ready to go data at the census tract level. Do pay attention to margins of error
+use,reac_scores,reac_scores_20170315,3/15/17,../../../data,https://s3.amazonaws.com/housinginsights/,raw/preservation_catalog/20170315/Reac_score.csv,
```
NealHumphrey: Let's pull this line out of the sample for the purposes of the dev version (it's a good place to user for testing, but should clean up afterwards)

python/scripts/meta.json
```diff
@@ -2084,5 +2084,54 @@
...
         ]
 
+    },
+
+    "reac_scores": {
+        "cleaner": "reac_score_cleaner",
+        "replace_table": true,
+        "fields": [
+            {
+                "display_name": "nlihc_id",
```
NealHumphrey: It would be good to edit the display name and display text etc. It looks like you've used display_text for the purpose of display_name. display_name should be the title displayed on a legend or other field (REAC Score) whereas the helper text should be a sentence or so describing what it is ("REAC scores come from HUD when they assess the physical condition of HUD related multifamily projects, typically once every few years per property.").

NealHumphrey: You can look up info from HUD on how they describe reac scores to give a useful summary - should give users enough context to at least know what to google to find more if they need to.

python/scripts/meta_sample.json
```diff
@@ -369,6 +369,55 @@
                 "type": "text"
             }
         ]
-    }
+    },
 
+
```
NealHumphrey: again, should remove this from the sample file - just add to the actual file

louvis:
@NealHumphrey I made changes to the files you mentioned and ran the autocrlf fix but it doesn't seem to have worked. I also realized that I hadn't merged the updated dev so I've got some more work to do to get this resolved.

NealHumphrey:
@louvis Rather than keep fighting Git, it's probably easiest to make a new branch off the current dev branch, manually copy-paste your changes in, and make a new pull request with that new branch. When it's not a big complicated branch that's often easier than resolving all the conflicts via Git. I recommend going that route - just make sure that after you do a `git add --all` run a `git status` and make sure that those .gitignore files aren't in there then - if they are it means you haven't fixed the line ending issue yet (can have someone help with that tomorrow night if not )

PR #190, opened 2017-04-11
WIP: Mockup (not quite demo) of declarative SQLAlchemy ORM syntax for tables

ohaaga:
First open source PR here. Still very much work in progress, but since I saw in mentioned in another post (@jasonrhaas) I figured I'd put it up before the meeting tonight. I've taken a few tables and turned them into declarative SQLAlchemy classes. With a bit of luck, this should allow us to have a single representation (in a single file, no less) of the metadata and parsing info for each dataset. As other people have pointed out, this representation is particularly convenient for writing APIs using flask or flask derivatives. I'm thinking we could also use the same representation for the data ingestion process, and mixins/Ingestible.py is my vision for how that might work. Unfortunately, when I try to demonstrate it using a quick sqlite db, I'm getting an error somewhere deep in the ORM, which I'm still trying to track down. This might not be a problem if someone more knowledgeable could hook this up to the Postgres database (I haven't tried). Other TODOs include transferring the cleaner code to the relevant classes using the @validates decorator, and putting some code into the Ingestible class to make sure the validations run properly when ingesting from files. Also coding in the relationships with the crosswalk.

Whether or not we use any of this code, I hope this helps to give everybody a picture of the costs/benefits of the declarative ORM syntax. See you tonight.


jasonrhaas:
Nice work!  It's a good example of how to use SQLAlchemy to simplify the ingestion process.  Have you looked at Flask-restless at all?  It has a built in `flask.ext.sqlalchemy` library that might simplify the database table creation even more.  It supports this and also "pure SQLAlchemy" which you have in this PR.

You can see the difference here:  http://flask-restless.readthedocs.io/en/stable/basicusage.html

ohaaga:
Thanks for looking over it, Jason.  I don't really understand a lot of the layers in the sqla session/engine/etc "pure" code, and I think the flask-sqla version you linked to looks a lot cleaner.  It may take me a while to figure out how to

* Split out the models into separate files without generating circular dependencies on the base class
* Run the ingestion code without necessarily firing up a flask app

I'd appreciate any tips if that's a case anybody has dealt with.

I figure my first step will be to add an indirection layer to the sqlalchemy imports so we can tweak the source of the base, column, and column type classes in one location.

NealHumphrey:
@ohaaga Great, thanks for doing this PR so we can all see what's going on w/ this declarative approach. This will give us a good starting point to discuss how this could integrate with / replace our ingestion code and the Flask api. 

I added 'WIP' to the front of the PR to more easily distinguish it as a work in progress. 

On a brief read through of the code it looks like it's on the right track. It is probably best to discuss w/ others who know sqlAlchemy / Flask / Flask-Restless better than I to give better direction on next steps. 

On the 'pure' sqlAlchemy differences between session/engine/connection that you note having issues with, I struggled a lot with this when starting sqlAlchemy. I ended up writing my summary as an answer to a stack overflow post when I finally got the hang of it because I couldn't find one good source to point to - you can [read that summary here](http://stackoverflow.com/questions/34322471/sqlalchemy-engine-connection-and-session-difference/42772654#42772654)

python/declarative/models/project.py
```diff
@@ -18,7 +18,7 @@ def parse_dates():
         raise NotImplementedError
 
     nlihc_id = Column(        
-        "nlihc_id",
+        "Nlihc_id",
```
NealHumphrey: Hey @ohaaga per your commit comment and this code, does this mean you're making the column name in the CSV file always match the column name in SQL? We want to be able to have those be different, actually. Being able to specify a different name solves a few problems for us:
- running SQL queries on fields with capital letters, spaces, etc. gets much more complicated - SQL allows you to create fields with names with non-standard characters but then requires you to always quote the field name in your query. This makes querying the database more difficult
- Sometimes the field name is not descriptive so we want a better name in our database
- Sometimes the same type of field is called one thing in one data source and something else in another - we want to be able to use the same name for all tables in our database. For example, in some data sets the neighborhood cluster is called 'neighborhoodcluster' and in others it is 'neighborhood_cluster' - we want all tables to use a consistent approach in our database.

NealHumphrey:
@ohaaga what do you think, do you want to just keep this branch for reference when doing the refactor but close this PR?

NealHumphrey:
Closing as @jasonrhaas is setting up this approach w/ our current code

Issue #200, opened 2017-04-14
WMATA upgrade api call


NealHumphrey:
There is currently code in `/python/scripts/wmata_distcalc.py` (from @prisalex) that pulls information from WMATA about available bus and metro stops and then calculates the walking distance from each building in our projects table to each metro/bus stop nearby. 

We want to update this to reflect our current approach to getting data from APIs, as well as fix a few implementation issues identified while ingesting the data. 

- [ ] Add a WmataApiConn class in `/python/housinginsights/sources`. Migrate existing methods to follow OOP approach
- [ ] Instead of a single `:` separated section in 'lines', create a new table with to columns: `station_code_or_stop_id` and `line` (i.e. normalize the data), This will make it easier for us to do queries like `select distinct line where station_code_or_stop_id in ['list','of','stops','near','building']` instead of reparsing the colon separated values every time. 
- [ ] Make sure column names are consistent both between tables (e.g. code_or_id vs. station_code_or_stop_id) and with other data sets (e.g. latitude instead of lat). 
- [ ] Update all the relevant stuff for adding these newly created files to the database (manifest,json,cleaner etc.) and removing the old data.

NealHumphrey:
From @nmolivo on Slack, adding here as a note about current status: "We are on the second bullet. We have a question on the following code. Do you mind walking through what it is doing?"  @NealHumphrey will coordinate with them - @jgordo04 and @nmolivo just let me know what specific question you have!

jgordo04:
Working on converting /python/scripts/wmata_distcalc.py to class based, but need to learn how to test it with a database. Have some hold ups with bullet 2 also, but should be cleared up with learning how to test it.

NealHumphrey:
@jgordo04 if #2 is tricky, I actually did already build the parsing into the API endpoint since it was needed for that - I'd suggest holding off on it for now and just flag it with a TODO and we can get to it at some point. It'd be better for advanced users that are tapping into the database to have this available, but it's not on a critical path.

PR #202, opened 2017-04-14
WMATA data and API endpoint

NealHumphrey:
- Adds the one-off version of the WMATA data tables (see #200 for next phase)
- Adds an API endpoint for nearby transit which returns both stops and routes that are accessible within a 0.5 mile walk.

### DIFF
--- a/docs/data/example.md
+++ b/docs/data/example.md
@@ -1,4 +1,5 @@
 ---
+---
 layout: datasource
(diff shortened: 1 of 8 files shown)


NealHumphrey:
Merging - tested locally and seems Ok, need to bring it in with the encoding param fix to get everything working properly. Would be good to get another set of eyes on it at some point though

Issue #193, opened 2017-04-12
MVP Additions ahead on demo Monday 4/17


ostermanj:
Here are the priorities coming out of tonight's work session for improvements to the MVP ahead of @NealHumphrey 's demo on Monday. It's all very related to the larger issue #143 .

1. Add ability to display **two fields** on top of the main map view: building permit count and crime count. These will be provided by API call, which should be ready for use by Wednesday evening. The API should return JSON (one array of objects) per zone aggregation type. Subtasks could be divided this way:

- [x] build menu, top-left of map, similar to zone selector that offers the two options of Crime and Building permit
- [x] selecting an option calls relevant API endpoint and returns appropriate data (or we fetch data on load in anticipation)
- [x] the harder part: have mapBox fill the zones based on the data

2. Link the selection of 'Neighborhood' to the options in the dropdown for the pie charts. (I can do this—didn't realize it was actually Neighborhood cluster, which has a one-to-one mapping to the project.csv data.

3. Add legend to map to explain the color-coding of the dots.

4. Leave the pie charts (NH edit: aggregating client side) ~~connected to static projects.csv file~~ (NH edit: now pulling from API) for now, but add (probably not programmatically—just force them in) the human readable field names as labels

5. Animate the pie charts when subzones are selected from the dropdown. I think this would be easier if the donuts were really showing just one arc ("yes") on top of a 360* background, rather than two arcs.

6. Some cleaning up of the building view page. Like can the be styled more like the map on the map view page? ~~Get rid of unused accordion sections?~~ (NH edit: should keep them we can fill in w/ some placeholders)

Let me know if anything's unclear, and let's let others know if you're jumping on anything.

NealHumphrey:
Hey @ostermanj and @ptgott, moving conversation to this ticket for now until we get the version for Monday deployed - we can do a sweep of issue #143 after Monday to pick up any lingering issues we haven't dealt with yet. 

- The API endpoints I described in #143 are live on the Flask API. 
    - Use the format `http://hiapidemo.us-east-1.elasticbeanstalk.com/api/raw/<tablename>` to get the raw data for any table from these currently available options: `['census_mapping', 'project', 'subsidy', 'acs_rent_lower', 'acs_rent_median', 'crime', 'acs_rent_upper', 'manifest', 'building_permits']`
[...]

PR #203, opened 2017-04-14
WIP: Building view improvements for demo

NealHumphrey:
- General page cleanup of CSS to use bootstrap properly and better lay out elements
- Added bus stops to the metro map (need to move this into transit section)


NealHumphrey:
I'm adding this pull request as a WIP so @ostermanj , @ptgott , and @nealfwilliams (and whoever else working on #193 ) can see what's going on with the building view side. If anyone wants to work on anything there let me know and we can merge in and coordinate who works on what sections. 

Next up for me is adding the WMATA data section using the API stuff I just put into #202 .

NealHumphrey:
@ptgott thanks! I've actually been working on  the API endpoints to do the 'x within n miles' so that I could add that to the view today (starting w/ metro since we have distances calculated already).

I agree on the building dots on zoom. I'd like to also figure out how to clean up the labeling since it makes it pretty muddy when the underlying map data interferes with building label. The only way to really do this is see if Mapbox lets us tap into their [labeling algorithm](https://www.mapbox.com/blog/placing-labels/) as avoiding label collision is a hard problem. 

My CSS is not great so I'll probably let someone else do accordion transitions

NealHumphrey:
It seems like the label collision avoidance should be occuring automatically - all the default settings in here seem to indicate that it will do the collision detection: https://www.mapbox.com/mapbox-gl-js/style-spec/#layers, not sure why it isn't doing it by default

NealHumphrey:
Merging this in so my API changes and the nlihc_id edits to the map view are available in dev.

PR #205, opened 2017-04-14
Create promise for map load to ensure app init function will always b…

nealfwilliams:
…e called.

Basically what this commit does is it creates a promise that will resolve when the map loads.  When you attach a callback to a promise via .then it will do one of two things.  If the promise has already been resolved it will run the registered function immediately.  Otherwise it will wait until the promise resolves, then call the function.


nealfwilliams:
I will fix conflicts then create a new pr

PR #208, opened 2017-04-15
Map load fix

nealfwilliams:
This is a pretty simple fix to the map load issue 206 that doesn't require promises or moving functions around in the code.


ptgott:
@nealfwilliams Thanks for this! Nice simple fix. I've removed the 'map_loaded' variable in favor of using Mapbox's own 'map.loaded()' method, then merged the PR.

PR #210, opened 2017-04-15
add data heatmap menu

nealfwilliams:
First pass of menu to toggle data heatmap.


NealHumphrey:
In Chrome this looks great!

I was in Firefox (43.01 then 47.0.2) and it didn't like the [forEach function](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach) in line 278 - "forEach is not a function", and didn't get the menu as @ptgott said.  After it upgraded itself to 52.0.2  (I haven't used Firefox in a bit), it no longer had the error. The docs (linked above) note that it doesn't have Firefox pre-50 support, and there is no Internet Explorer support.  @ptgott what version are you on?

ptgott:
@NealHumphrey I'm not at the computer I used for the review, but I should be up to date with Firefox. I'm not sure about Chrome. One cause of the issue on my end may have been the specific way in which I resolved the merge conflicts in my local test branch. Resolving the conflicts seemed straightforward at the time, but there's always the chance I missed something.

NealHumphrey:
I didn't merge, just checked out the branch locally, so yes that could also
be a difference.

nealfwilliams:
I merged the latest changes into my fork.  Everything is still working fine for me.  Resolving seemed straightforward to me as well, just keeping the code from issue #208.  

Regarding the issue @ptgott was experiencing, the only two lines in any JS file that reference the data_sources variable are 374-5 in rich-map.js.

var data_sources = document.getElementById('data-menu');
data_sources.appendChild(link);

So it seems like the only data_sources would be null is if there is no element on the DOM with id data-menu.  If more errors persist, could you double-check your local map-view.html file to ensure the data-menu div is there (line 37 at time of writing)?

Regarding the forEach issue, there was actually already a forEach call in the code before these changes.  I was mainly just trying to follow the patterns already being used.

Regarding the commented lines, this is where code would go to disable the data-source menu options on layers that do not have that data source.  That is what I was planning to work on next, but figured it's not that crucial for the demo.

NealHumphrey:
Thanks for the update @nealfwilliams . Sounds like it might have been a merge conflict resolution error. 
I closed @ptgott 's requested changes, so it can be merged now. I won't have a chance to test the updated version locally until this afternoon, but if @ostermanj or @ptgott do any work on this in the morning probably best to merge in first!

ptgott:
@nealfwilliams @NealHumphrey Works perfectly on my end. Thanks for your work on this! Merging now.

PR #211, opened 2017-04-15
Adds a shortest walking distance functionality for all transit routes…

NealHumphrey:
…, and adjusts the format out the output to match.

Slightly messy code format - would benefit from moving into functions. Needs unit testing.

This is deployed to live api


NealHumphrey:
Merging so we can use for demo. Should be revisited w/ testing and another set of eyes.

PR #214, opened 2017-04-16
Haversine (issue 213)

ohaaga:
Work in progress, but feel free to use as a starting point if you don't get any other PRs.  Unfortunately, I'm having a little trouble building the flask server, so this isn't even minimally tested.


NealHumphrey:
Thanks @ohaaga. I'm merging this into a different branch and will check that the endpoint works and add any additional changes. It looks like it should be pretty good!

Issue #176, opened 2017-04-06
Allow optional columns in the table_meta so that cleaners can add new data fields


NealHumphrey:
We want to be able to specify columns in the table_meta.json file that don't need to exist in the source data CSV file. These can either be a) optional fields not present in all versions of the data  or b) fields that do not exist in the source data but are instead added during the data cleaning process (such as geocoding fields). 

- [x] Add a 'required_in_source' param to each field in the make_draft_json function, with its default value set to False. 
- [x] Every table should have an additional field called `unique_data_id`. The json for unique_data_id should look like:
    ```
        {
                "display_name": "Unique data ID",
                "display_text": "Identifies which source file this record came from",
                "source_name": "",
                "sql_name": "unique_data_id",
                "type": "text"
                "required_in_source": False
            },
    ```
 The default 'required_in_source' param should set to False (since unique_data_id is added during the ingestion process)

- [x] Update the DataReader to be able to handle fields that do not exist in the source CSV file. The 'do_fields_match' validation will need to be updated to make sure that all required fields are there, and that no unknown fields exist. 
- [x] All Cleaner functions should be able to fill in missing values with the appropriate null value. This method will need to be called in every cleaner subclass. The existing replace_nulls method will not work because it only handles dict keys that exist in the `row` variable - new keys will need to be added to the row for missing values. 
- [ ] Create an appropriate TestCleaner 

### Background:
Currently, the HISql class in `python/housinginsights/ingestion/SqlWriter.py` initializes `self.fields` to match all the fields listed in the meta (being renamed to table_meta) variable. Then, it automatically adds a unique_data_id field. The create_table class therefore makes a table with all the columns in the table_meta plus a unique_data_id column. 

Meanwhile, the CSVWriter class (in `/ingestion/CSVWriter`) is used to take the rows of clean data provided by the Cleaner subclasses. It always adds a unique_data_id to the existing row dictionary. The 'write' method can probably stay the same (ensuring that every row of data has its associated unique_data_id) but the __init__ method will need to be modified so that it uses the table_meta version of the sql_field instead of manually adding it. 

Finally, the DataReader expects the number of rows to always match the table_meta; it enforces this by rejecting the table if the fields don't match (using the do_fields_match method). The SQLWriter requires all fields of the processed pipe-delimited file that is bulk-inserted into the SQL table to have values for every column of the database tables with an appropriate null for any missing value.

speedyturkey:
[...]

PR #216, opened 2017-04-16
Issue 176 optional columns

speedyturkey:
This PR fixes issue #176.

It is now possible to include optional fields in meta.json which do not exist in the source CSV. Rather than update the replace_nulls method, I thought it would be better to update the row with empty values before passing it to the cleaner - this way, the cleaner only has one job.

I noticed that not all subclassed cleaners call the replace_nulls method, which results in empty strings being loaded to the database rather than nulls. Given that clean is an abstract method in the base class, I wasn't sure how to force the subclasses to call replace_nulls but can research further if desired.

The various datareader classes have been updated to support these changes, with tests added for the do_fields_match functionality. I didn't add a test for the Cleaners because I didn't make any substantive changes there.

Additionally, as in my last PR, I've included both old and new meta.json files and the script used to generate the new one - please feel free to delete the old one as well as the script if the new one looks good.

As always, happy to discuss and grateful for feedback. Thanks!


NealHumphrey:
Great! I'll review this post-Monday, since we don't need it for the demo. Thanks @speedyturkey !

NealHumphrey:
@speedyturkey I'm very sorry I let this sit around so long! I got caught up in the rest of the planning and onboarding new folks. I will get this reviewed and merged in either tomorrow or Wednesday at the latest. Thanks for your work on this.

PR #217, opened 2017-04-16
Add permits and projects w/in distance

NealHumphrey:
Adds two endpoints to the API, one for nearby_projects and one for nearby_building_permits that use a lat/lon param to calculate the building permits within X miles. 

This also implements a couple changes to how jsonify writes json, to resolve some issues related to data types that this endpoint highlighted. 

This is deployed to the live site

TODO in a refactor we'll want to eliminate duplicative code

### DIFF
--- a/python/api/application.py
+++ b/python/api/application.py
@@ -7,7 +7,7 @@ This is a simple Flask applicationlication that creates SQL query endpoints.
 
 """
 
-from flask import Flask, jsonify, request, Response, abort, json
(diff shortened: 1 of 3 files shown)


NealHumphrey:
@ohaaga for reference here's the stuff I changed. Mostly it was issues related to data types that I had to sort out - params coming in as strings instead of int/decimal, sqlAlchemy row objects instead of dicts, etc. I also added the proper methodology for url params (request.args) and a few other features.

PR #219, opened 2017-04-17
WIP: Subsidy Timeline

NealHumphrey:
Various edits to get the timeline working in the building view page, but only with sample data

TODO - this needs to pull from this API endpoint instead: 
http://hiapidemo.us-east-1.elasticbeanstalk.com/api/project/<nlihc_id>/subsidies/


NealHumphrey:
@ostermanj or @ptgott if either of you are able to help me convert this to use the API instead of a flat example file, it's almost there! I'm just not sure best approach to get this to pull from the API

NealHumphrey:
Merging this in, I'm going to try to just drop the full static file in there and see what happens

PR #221, opened 2017-04-17
Animates pies on load and update; cluster names in the pie chart dropdowns

ostermanj:
This changes how the pies are drawn: one data-bound arc for % yes on top of a static background instead of two data-bound arcs. That makes the code simpler, especially for the transitioning (animating) the arcs when selecting a different zone type or zone.

Adds the neighborhood (cluster name) data to the drop down when the neighborhood map layer is selected. All but the Zillow layer are connected to the pie charts— @NealHumphrey for the demo beware that selecting Zillow does nothing to the pie charts. Disable it?

@nealfwilliams @ptgott

### DIFF
--- a/javascript/tool/css/map.css
+++ b/javascript/tool/css/map.css
@@ -198,3 +198,7 @@ p {
     margin-right: 5px;
     width: 10px;
 }
+
+#zone-selector {
+  max-width: 300px;
+}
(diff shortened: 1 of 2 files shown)


NealHumphrey:
This looks great! 

Zillow is fine to leave alone, though we may end up pulling it off anyways until we figure out how to deal with mapping zone-based data between different zones. For the heatmaps, only Ward and Neighborhood Cluster work, so I'll only be touching those in the demo - and we're not really releasing this to anyone to use outside of the demo environment yet.

Issue #212, opened 2017-04-15
JUNE: Front-end Refactor


NealHumphrey:
General catch all to put things that would be good to make more systematic in the next version. @ptgott and @ostermanj feel free to add stuff as you come across it. 

- Overall look at our code structure - how do we feel about it? Can we do some things more consistently? For example, function declarations as variables vs. the app dictionary (esp. across multiple files), IIFE vs. chaining of functions, promises etc.
    - Example: building view page split between building-header, building-map, etc.
- How should we handle legends. Currently a very hacky method in the 'nearby transit' section. Mapbox has SVG icons or custom-drawn circles. One icon has SVG raw copy-pasted into the building page, the other icon is generated via a tiny D3 svg infill that has duplicated properties as to what was put in mapbox. What's the best way to do this more systematically? Can we create a legend function? 
- Pull some stuff into functions, for example duplicated bus vs. rail code. 
- What's the best way to make api calls that depend on other API calls? For example, nearbyHousing and nearbyPermits in the building-map.js depend on having the buildingLat and Lon. Currently I solved this by making the whole thing in building-map.js a function variable which I call in building-header.js (this should be refactored a bit).

NealHumphrey:
Commenting on #2 - here's a demo of built in map-box legend, though we want to link ours out to other parts of the page too: https://www.mapbox.com/mapbox-gl-js/example/updating-choropleth/

ptgott:
Here's one thing we can talk about adding to the list: changing the division of labor between front-end and back-end so that the choice of datasets to fetch and display is done on the back-end. This way we won't have to, for example, manually enter information about each API query into rich-map.js. What this would allow would, in theory, be more extensibility in adding datasets (e.g. over 10 datasets to choose from on the data menu, each with aggregates for different zone types). A page would send a query for, say, 'map-view', and the server would know what datasets to return. The JS would then render the datasets appropriately. One way to do this, speaking generally, might be to include metadata in each dataset regarding what pages that dataset would be used for, then have the server return a big JSON object stringing together all datasets to return for a particular page. What do you think?

NealHumphrey:
@ptgott I think we probably don't want to totally push this onto the API, but I do think you're right that we need to build more flexibility into the front end and more uniformity into the back end, allowing the front end to render appropriately across different data sets using the same code. 

I'm concerned about future long-term performance implications of getting one big JSON object for each page. What if it is instead more user driven? Something like:
[...]

PR #226, opened 2017-04-24
Add a new directory for high-level refactoring options and discussion

ptgott:
#212 
I've also begun (like, super-preliminarily) to translate the whiteboard sketch into constructor declarations and .js files within the 'javascript/refactoring_sandboxes/option1' directory. 'javascript/refactoring_sandboxes/original_structure' takes you to a version of the dev branch's 'javascript/tool/js' directory in which everything but the main constructor, function and variable declarations have been taken out.

Let me know what's useful or not here and I can keep plugging away at it. 

I see this as more of a communication tool than an actual version of the code. For alternative structures, we could add other 'option' directories. It's less efficient than using git branches but lets you have the same file from each variation open in your editor.

### DIFF
--- a/javascript/refactoring_sandboxes/option1/chart_types.js
+++ b/javascript/refactoring_sandboxes/option1/chart_types.js
@@ -1,0 +1,24 @@
+// PG: It would be nice to reduce the number of arguments to something easier to
+// memorize. 
+// We still have to incorporate Bostock's guidance re: reusable charts.
+// We also have to work out how to integrate this with DataObject (see data_objects.js).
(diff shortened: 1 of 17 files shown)


ostermanj:
Thanks, @ptgott. Very useful to think through how to refactor. I keep coming back to a couple questions: should we be using a library to handle all the state, of which we have quite a bit? And should we build the tool as single page, whether or not we use a library? That would take care of the data-persistence problem from page to page.

Anyway, I'm thinking through an option 2, still in vanilla JS that may end up not being so different from option. Hope to push it tonight.

PR #230, opened 2017-04-26
WIP:  Flask restless api demo

ohaaga:
I've added a few files to python/api showing how we could use flask restless (per @jasonrhaas 's suggestion) to auto-generate the basic endpoints.  The magic can be seen by running

```
export FLASK_APP=../api/declarative_app.py
flask run
```

and maybe 
```
pip install -r ../api/requirements.txt
```

in python/housinginsights, then hitting /api/acs_median_rent, /api/projects or /api/census_mapping .

Ultimately I assume we want the models to each have their own file in the "models" folder.
Pain points in putting this together were:

* The default serialization in flask is a little anaemic.  I commented out a field in "projects" that was causing trouble.
* Had to mess around a bit with the join.  I don't like that it seems to require the database names of tables and fields rather than the python class and property names.  I hope I'm doing this wrong.

In general, I'm slowly overcoming my fear of SQLAlchemy internals.  This article helped a lot:
http://aosabook.org/en/sqlalchemy.html

If we're pretty sure we want to go in this direction, I'd like to look into using the PostGIS extension (and hopefully a SQLAlchemy wrapper on top of them) to allow us to do distance based queries directly rather than doing our own calcs in the endpoint code.


NealHumphrey:
@ohaaga and @jasonrhaas - I'm cleaning up pull requests and issues. What do you think, is it more beneficial to merge this in so you have it handy when doing the refactor? Or is it better to close it:? Something to discuss when you meet, let me know what you want me to do with it.

ohaaga:
@NealHumphrey can we keep this up for a minute?  @jasonrhaas It's running with the postgis image and the geographic extensions to sqlalchemy.  Adding graphene may be a better idea after we have some more meat on the sqlalchemy schema.  I'll hold off for the new all-docker setup before I mess with the ingestion scripts, but I'm guessing I'll have to make a small tweak to the metadata to declare the geo variables.  Let me know if you'd like to divide up the tables for the sqlalchemy schema.

jasonrhaas:
Looks good.  Are we sure we want to move over to the postgis image?  Are there any negative consequences to doing that?  Just want to make sure that we don't break anything in the process.

I actually haven't tested this -- Do I have to close @ohaaga's repo to test this out?  Usually I just switch to the branch and pull from there.

xee5ch:
I would presume it does not use the Python conda builds from continuum. I know you guys have done a good job of keeping the CPython pip deps and conda env deps in check but you will incur the confusion of running different Python VMs side by side potentially.

jasonrhaas:
@xee5ch For this demo, looks like he's just running it locally.  We can certainly run it in a docker container as well if we need to.  It could be run inside of the **sandbox** container, or better yet we could spin up a new container called **api** which runs the new Flask API and shares out a port to the local machine so it can be pinged from `localhost`.

ohaaga:
@jasonrhaas @xee5ch Sorry to be out of the loop for so long.  I should be able to work on this again starting tomorrow.  I've been running things locally, but I imagine that everything will eventually run inside Jason's docker sandbox (presumably on the anaconda distribution).  I haven't had much trouble in the past using pip and requirements.txt files along with conda, but if we're worried I can look into getting conda versions of the new dependencies and adding them to environment.yml.

ohaaga:
@NealHumphrey @jasonrhaas Sorry, some work stuff came up, and realistically I won't be able to do much for the next few weeks.  I'll be in touch if that changes, but probably can't do much for the initial launch.  Best of luck.

PR #231, opened 2017-04-30
Very preliminary steps toward setting up a test environment on the front end

ptgott:
#212 
Since we'll be moving around a lot of code during the refactor, it'd be great to have some tests set up for our front-end code. I've opted to use the standalone version of [jasmine](https://github.com/jasmine/jasmine) to avoid interfering with the back-end environment. The jasmine library lives in javascript/tool/jasmine-standalone-2.6.1. 

I haven't written any tests yet, as I'm still figuring jasmine out. The scant work I have done is to sketch out the beginnings of a structure for the tests file ('spec.js') as well as mention an SO answer page with a solution to an issue that comes with Mapbox: testing the JS when Mapbox GL JS looks for a particular HTML element. 

If you all are interested, we can talk about setting up a more fleshed out full-stack testing environment with things like Selenium, Karma, etc.


NealHumphrey:
@ptgott, never having used any JS testing environment, standalone jasmine sounds OK to me. I'm approving the PR but I will let you merge in either now or after you've written up a test.

ptgott:
@NealHumphrey 
I forgot to mention testing on Tuesday. I'm not going to have time to write tests this week, but after my involvement on the Browse side of things cools off I can switch to doing so (assuming the other teams are okay on the JS side of things!).

PR #232, opened 2017-05-01
Adds option 2 to sandboxing. proof of concept MVC pattern with PubSub

ostermanj:
This refactors the some of the bare-bone functionality of the existing tool into an MVC pattern, as an option and for discussion. Other functionality is placemarked.

The proof of concept initializes the mapBox instance and  publishes via PubSub when that style is loaded. Publishing that triggers a subscribed function to add the initial layers (ward, neighborhood, etc). If the needed data does not exist yet in the dataCollection (it doesn't), the data is fetched.

The addLayer function passes a callback to the getData function so that the layer is added only after that data is available. That helps sequence the asynchronous requests. 

It introduces a 'state' property to the model where crucial states of the app can be tracked. Next level could be to create a watcher function that publishes an event any time a property of the state changes. That could really decouple action from reaction and add flexibility. It may also mean creating from scratch a less-good version of any number of out-of-the box stateful libraries.


ostermanj:
Added a couple commits that add a StateModule that sets and gets the state and publishes via PubSub when a state change occurs.

With this approach, there would be no need to query the DOM to know the state of the app. That gets messy and perhaps slow as things get more complicated.

An action that would have an effect on the DOM or view need only set a new state property. That change is announced, and any existing or future function that needs to react can independently subscribe to that announcement.

I've missed the past two weeks of discussion, but however we decide to refactor I think this concept of abstracting the state away from the DOM and using a PubSub to  coordinate action/reaction could be a part of any solution short of embracing an Angular or React library.

ostermanj:
@ptgott @NealHumphrey @Janalyn @wassona 

Work in progress, about halfway toward reproducing all the features of the alpha version with an option for refactored code. Selecting the zillow layer still doesn't work with the pie charts, and we still need  a way to provide the readable labels that isn't hard-coded.

The chart constructors and inheritance are pretty much as they were. That refactoring is still to come, as necessary. The changes here center around a model-controller-view pattern for building the map view page.

# tl;dr

* fetch data with the `getData()` method which fetches data using d3.JSON if it is not already in the dataCollection. You can pass it a reference to a callback function to be run only when that data is ready. `getData([name: String[, params: Array[, callback: functionReference]]])` No need for setInterval checks on data being ready or to write our own AJAX calls.

* a StateModule keeps track of a manually constructed abstract state of the app and publishes, via PubSub, whenever a stateChange occurs so that other elements can react. It's not monitoring the DOM, so we should avoid changing the DOM without announcing the stateChange. `setState(property,value)`. For example, when the dropdown selection for the pie charts is changed, it does `setState('pieZone', e.target.selectedOptions[0].value);` rather than directly calling the function that changes the pie charts. This way other functions, perhaps not yet written, can easily hook into the same event.

* use PubSub to link the firing of one function to stateChange caused by another. Subscribe functions to PubSub announcements / topics with `setSubs()`. `setSubs([[topic0, functionReference0],[topic1, functionReference1], ... ])`. The code below, for example, subsbribes `sideBar.changeZoneType` to any pieZone stateChange (third array): `setSubs([
            ['firstPieReady', sideBar.setDropdownOptions],
            ['mapLayer',sideBar.changeZoneType],
            ['pieZone', sideBar.changeZoneType]
        ]);`

//tl;dr

# the dataCollection object

There is a dataCollection object of the model that houses all the data after it's been fetched from the API. I may modularize this with the getData method of the controller so that it's held as a private variable of the module, inaccessible to the outside except through the designated methods.

# the getData method

The getData method of the controller looks first to the dataCollection. If the requested data is not there, getData gets it from the API and puts it in the dataCollection.

getData takes one object as an argument with 'name' as the one required property (for instance: `{name: 'ward'}`). Optional properties (optional for the function but necessary for some data requests) are 'params' (an array) and 'callback' (a function reference).

The values in the params array, if present, become part of the url of the API call. With `{name: 'crime', params: ['all','ward']}`, for example, getData would append '/crime/all/ward/' to the API url, following the pattern of the existing API.

When data is added to the dataCollection, the controller updates the state of the app and that stateChange is broadcasted by PubSub; any functions subscribed to that stateChange would then fire. That could be the way to make sure mapLayers and charts and so on initiate only after data is available. For now, for those first initializations, I'm using another solution which set a callback function as a property of the object passed to getData.  getData is using d3.JSON to fetch data. If a callback has been passed to getData, it is called by d3.JSON's success function. There is no setInterval check for data being ready.

# the StateModule

The StateModule provides an object that tracks the state of the app and the methods to set, get, and log it. Only the controller has access but those methods are aliased into global namespace so you don't have to type so much. (Before deployment everything should be wrapped in IIFEs so that global won't be truly global, where it would be exposed to interference.) Any function or user event that has or may have an effect on other parts needs to set that state change (`setState(property, value)`). That change, if it is indeed a change, is broadcasted out by the PubSub so that anything that needs to react can.

This may seem overly complicated but there should be two main advantages. We, individual contributors, may not be aware of the functions that others or our future or past selves have made or will make that are supposed to be triggered by a change in state. With this method, we don't have to know, and we don't have to make direct connections between one function that triggers another. Instead, we announce that a change has been made, and any function already subscribed to the type of change is set to go, and any function yet to be written can easily subscribe to that change.

It also keeps us from having to query the DOM to know what the state of our app is, which can get messy and error-prone.

There is a potential fragility here. If we change something in the DOM without making a stateChange, the DOM and the state will get out of sync. But I think the advantages outweight that risk. With this approach, we would want to change the DOM only with functions that are reacting to an announced stateChange.

# the SubscribeModule 

This module keeps track of PubSub suscriptions and provides the methods for creating or canceling them in a way that doesn't allow duplication or overwriting. Any stateChange is automatically published/announced. Setting subscriptions is done through the setSubs method, which takes an array of topic-function pairs as an argument. 

For example

`setSubs([
            ['mapLoaded',mapView.addInitialLayers],
            ['mapLayer', mapView.showLayer],
            ['mapLoaded', sideBar.init]
        ]);`

means that mapView.addInitialLayers is called whenever the mapLoaded topic is published, and so on for the others.

ostermanj:
Latest commits rearrange how the PieCharts are made so that the specific donut charts inherit from a PieChart prototype, which inherits from the Chart prototype. Basically took everything that should be common to all pie charts and separated it out from everything that was specific to the donut charts on the first view.

Also, WIP toward initializing the overlay data and joining it to the geometries of the map.

ostermanj:
Also, I change the API calls to reflect changes to field names that took place on the data side of things.

NealHumphrey:
@ostermanj awesome, thanks for keeping on this! I can spend at least an hour tomorrow morning playing with this to understand what's going on so I can talk about it intelligently tomorrow. 

on your comment "we still need a way to provide the readable labels that isn't hard-coded" - my intent with meta.json (https://github.com/codefordc/housing-insights/blob/dev/python/scripts/meta.json) was to provide this mapping of data field to display name. We can serve this file from the API. We're mostly in a code-freeze during the API refactor, but I could add the endpoint for this if it would help figure this out.

NealHumphrey:
@ostermanj This is really good. Mostly got my head around it - I think it fits most of the challenges we've come across. As you point out in the comment, it solves the issue of individual contributors, and our past and future selves, not needing to know how things that are triggered based on a change in state connect. 

On @ptgott's comment re: waiting until functionality matches, I'd like to know how much further we can get in 1 more week and see if we can make the switch then. This will take people a bit of time to get used to, so we want people to get familiar with it as soon as possible. I suspect that with this as the basis, we can also have a couple other people contribute to the rest of the refactor if we're smart about allocating tasks. 

This week I finally made a diagram of the existing code for our data ingestion, we may want to make something similar to this to help train people on the MVC pattern: https://docs.google.com/drawings/u/1/d/1jVsE5cKm9O8n9HBZ-dCY8PO48R6L6VQFFhnIsKASgHY/edit?usp=drive_web

PR #233, opened 2017-05-02
194 dc housing api

jkwening:
Completed DCHousingApiConn class that pulls affordable housing data set from opendata.dc.gov as JSON object. You can add raw data file to /data/raw/tax_assessment/opendata/ when the python/cmd/data.py script is run with the follow command:

python python/cmd/data.py DCHousing get_json --outtype 'csv' --output data/raw/tax_assessment/opendata/out.csv

Notes:
I had some issues with ImportError while trying to test and follow MarApi example template. I did the following in order to resolve those issues:
- export housing-insight into my PYTHONPATH: in order to get data.py to run
- I had to prepend 'python' for my module imports


python/cmd/data.py
```diff
@@ -42,9 +42,12 @@
 from argparse import ArgumentParser
 import importlib
 
-from housinginsights.config.base import HousingInsightsConfig
+import sys, os
+sys.path.append(os.path.abspath('../'))
```
NealHumphrey: If you change this to `sys.path.append(os.path.abspath('../../'))` you can omit the python in front of the import statements (e.g. `just from housinginsights.config.base import HousingInsightsConfig)(

python/cmd/data.py
```diff
@@ -42,9 +42,12 @@
...
 
-API_MODULE = 'housinginsights.sources'
+from python.housinginsights.config.base import HousingInsightsConfig
```
NealHumphrey: see above for how to remove the word python.

jkwening:
I actually added a few more lines to circumvent the hard coded path in `sys.path.append(os.path.abspath('../../'))` so I could successfully call the module irrespective of where I was in the housing-insights project folder. This means you don't have to navigate down to python/cmd folder in order to run the module - you can run it anywhere from the project folder. That was what causing me issues; I was always trying to call the module directly from the housing-insights folder. Let me know what you think. Thanks.

python/cmd/data.py
```diff
@@ -42,6 +42,23 @@
...
+while os.path.basename(os.path.abspath(path)) not in ["housing-insights",
+                                                      "python"]:
+    path += "../"
+
+# if at housing-insights folder, move down to housing-insights/python
+if os.path.basename(os.path.abspath(path)) == "housing-insights":
+    path += "python/"
+
+sys.path.append(os.path.abspath(path))
+
```
NealHumphrey: So here's some code I pulled from another project - a one-liner that you can use to always get the directory path of the repository, which you can then join to whatever relative path you need. 

`repo_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir, os.pardir))`

You'll need to edit the number of os.pardir as necessary. But, using __file__ instead of os.path.basename means it uses this file's path, not the path it was run from. 

I'll probably try to make a universal solution for this and go in and edit it in all the script files so it's done the same everywhere - if this is working, no need to edit for now.

NealHumphrey:
@jkwening - I tested and works for me, with the exception of the folder issue I noted in the review. If you can make that fix we can merge in. If you're comfortable w/ class inheritance, it would also be good to do the switch to use the BaseApiConn instead (and I'll edit the MarApiConn to do the same), and ideally hoist that results_to_csv function up a level. Let me know if that makes sense or if you need help on that.

jkwening:
The following refactors were completed:

- hoist 'result_to_csv' to BaseApiConn
- inherit from BaseApiConn instead (Note - refactored mar.py similarly)
- implemented recommended online-liner for folder issue. I will continue to update other files as I come across them.

Verified DCHousing.py using the following command-line with housing-insight project folder as cwd:
`python python/cmd/data.py DCHousing get_json --output data/raw/tax_assessment/opendata/refactor.csv`

I think this is good to go now. Let me know if I missed anything.

jkwening:
@NealHumphrey okay, this is good to go now. It includes all refactoring that I did to debug the 'unique_data_id' issue and standardizing with relative paths. Took me longer than I thought because I wanted to re-run everything to make sure I didn't break anything. It went through the load process without a hitch, so ready for PR.

PR #234, opened 2017-05-02
WIP: 212 sandbox option 3

NealHumphrey:
This is not really a working demo, but I tried working on it so figured I'd push it up. Mostly I spent time fighting with Javascript over how to handle protypical inheritance (I'm used to Python classes so kept expecting to be able to treat some of this the same). 

@ostermanj and @ptgott FYI for discussion


ptgott:
Since I've merged #232 as 'option 2', shall we make this an 'option 3'? How would you describe its relationship to the other two options (where option 1 is the super-rough sketch based on the whiteboard conversation and option 2 is the mostly working MVC pattern)?

NealHumphrey:
@ptgott I moved this into option 3. I'm not sure how much is worth using from this, but we could have it as an option until we finalize approach. This one was intended to be a clean from-scratch implementation of the whiteboard sketch with a 'hello world' example. I hit a couple JS challenges so didn't get as far as I'd hoped.

NealHumphrey:
Closing in favor of @ostermanj 's way better method

Issue #133, opened 2017-03-17
opendata.dc.gov files have extra bytes at file start


NealHumphrey:
Reading from the beginning of the file (i.e. first header row) of files downloaded from opendata.dc.gov produces this type of thing:

"\u00ef\u00bb\u00bfobjectid"

This is true for both the dc_tax data and the building_permits data. Telling Python to parse as utf8 (vs. latin1) produces similar results, though slightly fewer extra byte characters. 

For now I am copy-pasting all the contents of the file from a text editor into a brand new file to resolve this. Will need automated method in the future, probably by figuring out what the encoding type is, or parsing those extraneous bytes out. 

We need the DataReader class to be modified to be able to handle this.

NealHumphrey:
This is probably BOM headers:
https://en.wikipedia.org/wiki/Byte_order_mark

dogtoothdigital:
Hey there; I've been experimenting with this issue and I've figured out a potential solution by amending the `DataReader` class as you've requested.  I tested it against a few of the opendata.dc.gov files you've mentioned and the fix seems to work correctly.  I'd like to test it more thoroughly and review everything well before I submit a pull request.

`(housing-insights) ubuntu@ip-[redacted]:~/projs/hi/housing-insights/python/scripts$ python load_data_bomfix.py docker rebuild`
`meta.json imported. JSON format is valid: True`
`Manifest table created in the SQL database`
`sql_manifest_exists: True`
`building_permits_2017: preparing to load row 2 from the manifest`
`  File not found. Attempting to download file to disk:` `https://opendata.arcgis.com/datasets/81a359c031464c53af6230338dbc848e_37.csv`
`  Unknown content length; proceeding anyways...`
`  Download complete.  430679 bytes downloaded.`
**`  Leading BOM (byte order mark) found in file '81a359c031464c53af6230338dbc848e_37.csv'; removing...`
`  do_fields_match: True. meta.json and csv field lists match completely for 'building_permits'`**
`  Cleaning...`
`  Loading...`
`  replacing existing table`
`  building_permits table can't be dropped because it doesn't exist`
`  Table created: building_permits`
`  opening /home/ubuntu/projs/hi/housing-insights/python/logs/temp_building_permits_2017.psv`
`  data file loaded into database`
.
.
.

[...]

PR #242, opened 2017-05-05
133 bom removal in opendata dc gov file headers

dogtoothdigital:
Modified `DataReader` class to handle byte order marks (BOMs) in data file headers, currently only found in opendata.dc.gov-sourced files (actually served from opendata.arcgis.com).  Ran tests on the dc_tax and building_permits data files.  (Ref: [Issue #133 comment](https://github.com/codefordc/housing-insights/issues/133#issuecomment-299519507))

After this merge, all opendata.dc.gov data files with BOMs no longer need to have the byte order mark manually stripped before uploading to the AWS S3 bucket.

Ref changes in `DataReader` class:

[Modifications to `download_data_file(...)`](https://github.com/dogtoothdigital/housing-insights/blob/dev/python/housinginsights/ingestion/DataReader.py#L198-L213)

[Addition of `_remove_bom_from_file(...)` and `_contains_leading_bom(...)`](https://github.com/dogtoothdigital/housing-insights/blob/dev/python/housinginsights/ingestion/DataReader.py#L294-L321)

-meg

### DIFF
--- a/python/housinginsights/ingestion/DataReader.py
+++ b/python/housinginsights/ingestion/DataReader.py
@@ -8,10 +8,13 @@ import csv
 from os import path
 import os
 
-import logging
+import sys, logging, traceback
 
-from urllib.request import urlretrieve
-from urllib.request import urlopen
+from http.client import HTTPConnection, HTTPSConnection, HTTPResponse, HTTPException, OK
+from urllib.parse import urlparse
+from urllib.request import urlopen, urlretrieve
+from urllib.error import URLError
+
 import codecs
 
 from datetime import datetime
@@ -32,10 +35,10 @@ class HIReader(object):
     Note, local files are preferred when possible for faster processing time
     and lower bandwidth usage.
     """
-    def __init__(self, path, path_type="file", encoding="latin-1"):
+    def __init__(self, path, path_type="file", encoding="latin-1", keys=None):
         self.path = path
         self._length = None
-        self._keys = None
+        self._keys = keys
         self.path_type = path_type
         self.encoding = encoding
 
@@ -101,7 +104,7 @@ class HIReader(object):
 class ManifestReader(HIReader):
     """
     Adds extra functions specific to manifest.csv. This is the class that
-    should be used to read the manifest and return it row-by-row. 
+    should be used to read the manifest and return it row-by-row.
     """
 
     _include_flags_positive = ['use']
@@ -127,7 +130,7 @@ class ManifestReader(HIReader):
                     row['data_date'] = datetime.strftime(_date, '%Y-%m-%d')
                 except ValueError:
                     row['data_date'] = 'Null'
-                
+
                 #return the row
                 yield row
 
@@ -192,33 +195,127 @@ class DataReader(HIReader):
         self.manifest_row = manifest_row
         self.destination_table = manifest_row['destination_table']
 
-        # use default encoding if none found
+        # Use default encoding if none found
         if 'encoding' not in manifest_row:
-            logging.warning("  Warning: encoding not found in manifest. "
+            logging.warning("  Warning: encoding not found in manifest. " \
                             "Falling back to latin-1.")
         self.encoding = manifest_row.get('encoding', 'latin-1')
 
         self.load_from = load_from
-        self.s3_path = os.path.join(manifest_row['s3_folder'],
-                                    manifest_row['filepath'].strip(
-                                        "\/")).replace("\\","/")
+        self.s3_path = os.path.join( manifest_row['s3_folder'],
+                                     manifest_row['filepath'].strip("\/")
+                                        ).replace("\\","/")
+        self._error_reporting_overhead = {}
+        # Test connection to s3 URL
+        if self.manifest_row['include_flag'] == 'use':
+            self._suppress_verbose_error_reporting()
+            content_type_header = self._test_connection_to_URL(self.s3_path)
+            self._restore_verbose_error_reporting()
+            if content_type_header is not None:
+                logging.info("  Content (MIME) type: {}".format(content_type_header))
 
-        # load data file from given file system - S3 or local
+        self._keys = None
+        # Load data file from given file system - S3 or local
         if load_from == "s3":
             self.path = self.s3_path
             self.path_type = "url"
         else:
-            self.path = os.path.join(path.dirname(__file__),
-                                     manifest_row['local_folder'],
-                                     manifest_row['filepath'].strip("\/"))
+            self.path = os.path.normpath(
+                            os.path.join(path.dirname(__file__),
+                                         manifest_row['local_folder'],
+                                         manifest_row['filepath']))
             self.path_type = "file"
 
             if self.manifest_row['include_flag'] == 'use':
-                self._download_data_file()
+                headers = self._download_data_file()
+
+        self._keys = self._set_keys()
 
         self.not_found = [] #Used to log missing fields compared to meta data
 
-        super().__init__(self.path, self.path_type, self.encoding)
+        super().__init__(self.path, self.path_type, self.encoding, self._keys)
+
+    def __iter__(self):
+        """
+        Iterator wrapper that only uses canonical header fieldname keys
+        provided by _set_keys().
+        (Note: These canonical keys differ from those yielded by DictReader's
+               fieldnames property whenever the header row contains a BOM.)
+        """
+        self._length = 0
+        self._counter = Counter()
+
+        if self.path_type == "file":
+            with open(self.path, 'r', newline='',
+                      encoding=self.encoding) as data:
+                # Always use canonical header fieldname keys:
+                reader = DictReader(data, fieldnames=self._keys)
+                header = next(reader)   # Omit header row from output
+                for row in reader:
+                    self._length += 1
+                    yield row
+
+        elif self.path_type == "url":
+            with urlopen(self.path) as ftpstream:
+                # Always use canonical header fieldname keys:
+                reader = DictReader(codecs.iterdecode(ftpstream, self.encoding),
+                                    self._keys)
+                header = next(reader)   # Omit header row from output
+                for row in reader:
+                    self._length += 1
+                    yield row
+
+        else:
+            raise ValueError("Need a path_type")
+
+    @property
+    def keys(self):
+        return self._keys
+
+
+    def _download_data_file(self):
+        """
+        Internal function that tries to load the data file from the local file
+        system. If the file isn't found, it then downloads it.
+
+        :return: data file headers
+        """
+        headers = None
+        try:
+            with open(self.path, 'r', newline='', encoding=self.encoding) as f:
+                logging.info("  Defaulting to local data file: '{}'".format(self.path))
+                myreader = csv.reader(f, delimiter=',')
+                headers = next(myreader)
+        except FileNotFoundError as e:
+            self._validate_or_create_path()
+            logging.info("  File not found. Attempting to download file to disk: " +
+                         self.s3_path)
+            try:
+                urlretrieve(self.s3_path, self.path)
+                file_size = 0
+                if os.path.isfile(self.path):
+                    file_size = int(os.stat(self.path).st_size)
+                if file_size > 0:
+                    logging.info("  Download complete.  " \
+                                 "{} bytes downloaded.".format(file_size))
+                else:
+                    logging.warning("  Warning: Download cannot be verified; " \
+                                    "check location: '{}'".format(self.path))
+                    return
+                with open(self.path, 'r', newline='', encoding=self.encoding) as f:
+                    myreader = csv.reader(f,delimiter=',')
+                    headers = next(myreader)
+            except URLError as ue:
+                logging.error("  Error: _download_data_file(): " \
+                              "Error opening URL '{}' for download.".format(self.path))
+                raise
+            except OSError as oe:
+                logging.error("  Error: _download_data_file(): " \
+                              "Downloaded file '{}' cannot be opened " \
+                              "for reading.".format(oe.filename))
+                raise
+
+        return headers
 
     def _validate_or_create_path(self):
         """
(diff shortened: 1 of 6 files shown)


dogtoothdigital:
Hey, Neal, I know this isn't the highest priority issue / PR by any means but I think it's good to go.  If you'd like me to make any changes or have any questions, please let me know.

Cheers, 
-meg

dogtoothdigital:
Hey, Neal -- I know this isn't the highest-priority issue by any means but I think it's good to go.  Please let me know if you have any questions, concerns, or requests.
Cheers, 
-meg

NealHumphrey:
@dogtoothdigital Meg this looks like it works well, and in your comments on the issue it looks like you tested it well. I think the core functionality is good to go. 

I do think we should move where the call to `_remove_bom` occurs, though. There's a few issues with only calling it during the download phase:
1) When someone first gets a new data source that has a BOM, they will try to ingest it locally before pushing to S3. This is where the current code has problems with BOM headers, so this will crash and they won't get past this phase. 
2) Since we're (currently) using the AWS CLI Sync tool, if the file is modified locally and saved to disk, then the user adds a new file locally and syncs it will upload this modified file (without BOM). That will overwrite the original (non-BOM) version. Not totally a problem but a) we'll be modifying the original data (hoping to avoid this once we get into production) and b) it'll cause a bit of confusion for users. 

Is it possible to instead strip the BOM header in the `__iter__` method of the HIReader class?

option 1) Change `_remove_bom_from_file` to make a copy of the file and tell __iter__ to use the copy instead. To minimize code changes you could add an 'inplace=False' param to this method and just use an if statement to change which file is written to (same file or new file). 

option 2) Trim the relevant bytes off the file stream before the DictReader gets it. Not sure how hard this is. 

Let me know if you aren't sure how to do this, or if you think there is a different alternative that can solve the issues noted above.

dogtoothdigital:
Hey, Neal--

Yes, there's definitely a way to accomplish BOM-handling less obtrusively while reading the file stream without changing the data file itself.  Basically just had to set the class's header keys explicitly minus the BOM in the leading header field name (if detected) in the constructor, then overloaded/adapted `__iter__()` function and `keys` property accordingly and also handled latin-1-encoded BOMs directly without converting to utf-8.  I've got it working; just want to test & review it to make sure it's all good.  Re `HIReader` -- could apply the changes to the base class but was trying to restrict changes as much as possible to `DataReader` because the BOM issue shouldn't affect other *Readers that don't handle external data files.

cheers, 
--meg

NealHumphrey:
@dogtoothdigital Ok - go ahead and put it on the DataReader class then; if we ever need it elsewhere we can hoist it then. 

Thanks for taking another crack at it, let me know when it's ready for review again!

dogtoothdigital:
@NealHumphrey

Finally got around to wrapping this PR up(!) The following is ready to review:

#### New IO-stream-based BOM-handling in housinginsights.ingestion.DataReader that leaves data source files intact:
[`python/housinginsights/ingestion/DataReader.py`](https://github.com/codefordc/housing-insights/compare/dev...dogtoothdigital:133-bom-removal-in-opendata-dc-gov-file-headers#diff-65af713014abdf978117a4cf13869891)

Ref:
- `HIReader` and `DataReader` constructors
- `DataReader`'s overriden `__iter__()` method and `keys()` property
- `_download_data_file()` and `_set_keys()`
- `_contains_leading_bom(...)` and `_remove_leading_bom(...)`

[`python/housinginsights/ingestion/Cleaners.py`](https://github.com/codefordc/housing-insights/compare/dev...dogtoothdigital:133-bom-removal-in-opendata-dc-gov-file-headers#diff-9db39acd0ed61ecb8729b4807f0e04bf)
- `DCTaxCleaner` added to support handling dc_tax data with `load_data.py`
(Updated the corresponding `cleaner` reference in [`meta.json`](https://github.com/codefordc/housing-insights/compare/dev...dogtoothdigital:133-bom-removal-in-opendata-dc-gov-file-headers#diff-897e9e4577cd612c67146f6c45ddac7e))

#### Unit Testing for the BOM-handling version of `DataReader.py`:
[`python/tests/test_DataReader_bom_handling.py`](https://github.com/codefordc/housing-insights/compare/dev...dogtoothdigital:133-bom-removal-in-opendata-dc-gov-file-headers#diff-7ee9aa62c3f726f643ba368dc9929b4a)

See test code for cases:
- "Case 1: Test sample data file with bom in header row"
- "Case 1a: `load_from='file'`: Sample data file does not already exist locally and DataReader downloads remote file"
- "Case 1b: `load_from='file'`: Sample data file already exists locally and DataReader reads from local file"
- "Case 1c: `load_from='s3'`: DataReader reads from remote file"
- "Case 2: Test sample data file without bom in header row"
- "Case 2a: `load_from='file'`: Sample data file does not already exist locally and DataReader downloads remote file"
- "Case 2b: `load_from='file'`: Sample data file already exists locally and DataReader reads from local file"
- "Case 2c: `load_from='s3'`: DataReader reads from remote file"

...supported by test versions of `manifest.csv` and `meta.json`:
[`python/tests/test_data/manifest_bom_handling.csv`](https://github.com/codefordc/housing-insights/compare/dev...dogtoothdigital:133-bom-removal-in-opendata-dc-gov-file-headers#diff-b3635514a5c95e82b37cbe199413ab8f)
[`python/tests/test_data/meta_bom_handling.json`](https://github.com/codefordc/housing-insights/compare/dev...dogtoothdigital:133-bom-removal-in-opendata-dc-gov-file-headers#diff-3a98d3176d786514aff805e8749e837b)

Let me know if you have any questions/comments/requests...
--meg

NealHumphrey:
Also, yes I can squash and merge whenever it's ready to go in, so don't worry about extra commits.

dogtoothdigital:
Hey, Neal--

Thank you for the review + detailed feedback.  Wow, damn, I'm sorry the bom-handling tests are bombing on your end; I'll try to get to the b(ott)om of this ASAP!
;p

-meg

NealHumphrey:
Sure thing! 

For context, this issue became more important b/c we found how to get the BOM-ed CSV's directly from the opendata.dc.gov API so we will need to handle them in production. So glad you're handling it so thoroughly!

dogtoothdigital:
Hey, Neal--

Still probing this issue trying to replicate the unit test + `load_data.py` failures you encountered.  Here's an update:


(1) A note about how `test_DataReader_bom_handling.py` currently identifies which files are supposed to contain leading byte order marks:

- The way the testing in `test_DataReader_bom_handling.py` is set up right now, the testing method uses the `unique_data_id` in the testing/sample manifest file to identify whether the referenced data source has a leading byte order mark or not. That way it can test DataReader (via logged output and `__iter__()` data-row-handling functionality) for expected behavior in the case of BOM-led and non-BOM headers alike.  **Specifically, BOM-led files must include the substring `"_bom_"` in the `unique_data_id` column whereas BOM-free files include `"_nobom_"`.**

(I neglected to include a comment to this effect in the top of the unit testing file; I apologize for not making this clear!)

* * *

(2) Fixed an issue in `test_DataReader_bom_handling.py` w/r/t encoding that your full-length sample input file triggered.

- Issue: Whenever a stray 'exotic' latin-1 (a/k/a iso-8859-1) character appears anywhere in an input data file (which is rarely, but does occur in the `Integrated_Tax_System_Public_Extract_Facts.csv` file on line 103554, cols 88-89), the test would fail because the test code wasn't reading the source data file in the default latin-1/iso-8859-1 encoding when it should in order to test each source data file row against the corresponding DataReader-provided data row.

Ref:  (Warning, more down-in-the-weeds byte-level file content follows, aaaggggghh...)

`[redacted]@[redacted]:/mnt/c/[redacted]/Dev - Housing Insights/DataStudy/test_issue_133$ \`
`> head -n 103554 Integrated_Tax_System_Public_Extract_Facts.csv | tail -n 1 | hexdump -c -n 96`
`0000000   1   0   3   5   5   3   ,   2   5   8   2                   2`
`0000010   0   0   6   ,   ,   0   1   6   ,   R   e   s   i   d   e   n`
`0000020   t   i   a   l   -   C   o   n   d   o   -   H   o   r   i   z`
`0000030   o   n   t   a   l   ,   7   7   0   ,   1   7   0   6       -`
`0000040       1   7   1   0       S   U   M   M   I   T       P   L`
`0000050   N   W       1   7   1   0   �   �   ,   0   3   6   ,   M   O`
`0000060`

`[redacted]@[redacted]:/mnt/c/[redacted]/Dev - Housing Insights/DataStudy/test_issue_133$ \`
`> head -n 103554 Integrated_Tax_System_Public_Extract_Facts.csv | tail -n 1 | hexdump -C -n 96`
`00000000  31 30 33 35 35 33 2c 32  35 38 32 20 20 20 20 32  |103553,2582    2|`
`00000010  30 30 36 2c 2c 30 31 36  2c 52 65 73 69 64 65 6e  |006,,016,Residen|`
`00000020  74 69 61 6c 2d 43 6f 6e  64 6f 2d 48 6f 72 69 7a  |tial-Condo-Horiz|`
`00000030  6f 6e 74 61 6c 2c 37 37  30 2c 31 37 30 36 20 2d  |ontal,770,1706 -|`
`00000040  20 31 37 31 30 20 53 55  4d 4d 49 54 20 50 4c 20  | 1710 SUMMIT PL |`
`00000050  4e 57 20 31 37 31 30 c2  a0 2c 30 33 36 2c 4d 4f  |NW 1710..,036,MO|`
`00000060`
----------------------------------------^  ^
(Note: `0xc2` represents iso-8859-1 'Â' and `0xa0` represents the non-breaking space character.
These appear to be stray/erroneous characters in this particular file; perhaps cleaners should screen for such instances?  Might be tricky though because you don't really want to disallow or discriminate against international terms/names with diacritics and all that.)

* * *

3. Ran improved `test_DataReader_bom_handling.py` unit test with the addition of your realistic (hell, I mean real & unexpurgated) test sample input dc_tax file (`Integrated_Tax_System_Public_Extract_Facts.csv`) :

(_N.B._: Updated `manifest_bom_handling.csv` and `meta_bom_handling.json` under `python/tests/test_data` accordingly as per the condition described above w/r/t unique_data_id flagging of bom-led vs. non-bom input files in the manifest.)



`(housing-insights) ubuntu@ip-[redacted]:~/projs/hi/housing-insights/python/tests$ \`
`> python test_DataReader_bom_handling.py`
`Performing setUp() preparation...`
`Running test_bom_handling()...`
`Performing tearDown() cleanup...`
`.`
`----------------------------------------------------------------------`
`Ran 1 test in 109.297s`

`OK`

Re "1 test" which is actually 6 separate test (sub)cases -- yeah, I admittedly ought to refactor that behemoth unit test method, sorry!  And as for running for almost two minutes, that's because the `Integrated_Tax_System_Public_Extract_Facts.csv` is a rather large mofo and the line-by-line `assertEqual(...)` comparing each stinkin' input file data row against the corresponding DataReader-yielded data row takes a !$@#ing while on a large-ish file.

* * *

4. Now back to `load_data.py` for more testing on the same sample test data set...

(Slapping the results of this particular adventure in a gist thingamajig; stay tuned...
The tl;dr is that I cannot replicate the failure on your end thus far and I'm not sure why...  I've been performing all test runs in an aws ec2 ubuntu linux instance so that may be one difference in contexts/setups in case you're using docker on windows; I'll try to think of other potential differences in your dev sandbox vs. mine that could trigger failures(s)...  Because the code works, dammitiswear...  Although if I'm wrong and you've identified (a) clear problem(s), then I definitely want to revise it to robustly perform regardless of dev sandbox conditions, etc.)

-meg

dogtoothdigital:
Here's more proof that I can't yet replicate the failures...  More head-scratching and cogitating and whatnot is necessary; gonna take a smoke break and think a bit more...

["Investigating load_data.py performance for PR #242, Issue #133, continued..."](https://gist.github.com/dogtoothdigital/48bd54f0f4f8662e05ec3356a4b282e3)

NealHumphrey:
Thanks again for all this! This has again confirmed my disdain for all things encoding related. 

Were you able to replicate either of the issues? Or just not the unittest one? 

If you push your latest code up on this branch, I can try running it on Docker vs. my local python install and see if that changes anything, and dig into it some more in the morning.

dogtoothdigital:
Yes, I just managed to replicate it by retracing what it looked like you did as closely as possible; looking into it right now.  It is indeed related to the underlying stream's encoding; working on trying to resolve it.

Btw, is there any strong reason we default to 'latin-1'/'iso-8859-1' vs. 'utf-8'?  Is it tied to the data files themselves, retrieval via the HTTP protocol, or ... ?  Right now we don't detect the file's encoding as it's just stated in the manifest but I was wondering if maybe it might be better to try to detect it.

cheers, 
-meg

NealHumphrey:
The only reason to default to Latin-1 is that it works more often - UTF-8 stuff can be read as Latin-1 successfully more frequently than the other way around. We usually don't know what the encoding is at all, so any of those that are listed as Latin-1 in manifest.csv are fine to change to utf if it makes sense. If it's not too hard, adding sniffing in the make_draft_json would be a welcome addition. [relevant package](https://pypi.python.org/pypi/chardet)

In fact, just the presence of a BOM means that it's almost certainly UTF-8 encoding. So, we should probably change the encoding of anything downloaded from opendata.dc.gov to be UTF-8 and try your code with those settings.

NealHumphrey:
@dogtoothdigital Can you also push your latest code to the same branch so it appears on this PR? That way I can test w/ your latest fixes.

NealHumphrey:
Hey @dogtoothdigital on Tuesday it said it sounded like you got all this working? Can you push your commits up here so I can test the new version locally? I'd love to get this merged into the codebase before the hackathon Saturday if possible.

dogtoothdigital:
Hey, @NealHumphrey, just pushed the commits.

Sorry about the delay!

-meg

python/housinginsights/ingestion/DataReader.py
```diff
@@ -363,3 +456,57 @@ def _verify_required_metadata_columns_exist_in_actual(self, field_list):
...
+                                                'latin-1' also supported)
+        :type encoding: str
+        """
+        if (encoding == 'utf-8'):
+            bom_as_hex = "efbbbf"   # hex encoding of utf-8 BOM
+            str_as_hex = "".join("{:02x}".format(c) for c in str.encode())
+            if str_as_hex.startswith(bom_as_hex):
+                return True
+        elif (encoding == 'latin-1'):
+            bom_in_latin_1 = "ï»¿"  # latin-1-encoded version of BOM
```
dogtoothdigital: This should work now; DataReader now consistently passes the encoding through to all encoding-sensitive methods.

NealHumphrey:
Thanks @dogtoothdigital! I ran this with the tax example that errored out previously and it now works. Great!

I'm having some issues running tests but I think most of those are related to other stuff. I'm going to merge in now and then try to sort out tests from the dev branch.

Issue #183, opened 2017-04-07
Run everything inside docker-compose


jasonrhaas:
Python environments are notoriously difficult to get set up, especially when dealing with data science packages.  Anaconda helps with this quite a bit, but I've found there can still be problems depending on the user's specific local environment.  Docker of course solves all of this, which is why I've started running everything that I can inside a docker-compose environment.

We could create a Python sandbox Docker image that is used to connect to the database, and has all the connections set up already so the user doesn't have to deal with figuring this out.  Docker-compose by default shares all the container ports with each other, so we don't have as many local port conflicts either.

We load the data from inside the Python Docker container, which runs the Conda environment.  This way there is no need for a user to install Conda or any packages on their local machine.

Something like this would work and just exist to run commands on:

```
    sandbox:
      build:
        context: .
        dockerfile: Dockerfile-sandbox
      volumes:
        - .:/code
      command: tail -f /dev/null
```

All of the code is mounted, so no need to worry about re-building the container on code changes, unless there is a new package requirement.

rschmidtz:
I totally agree with this, I run out with dependencies problems when I try to set up my environment from requirements.txt with virtualenv (I use Ubuntu Linux). 
It would be easier to keep all the python packages and dependencies in a Docker container

NealHumphrey:
@jasonrhaas and @rschmidtz I definitely agree too. It would be great to have this bundled into a single docker-compose that we can run with `docker-compose up -d` so that it starts the other services and then makes the command line available with those services running. 

The reason this didn't make it into Tuesday's version is that I ran into issues getting the Housing Insights code to run in the container. I was using this docker image that came from the Code for DC Project Management project: https://github.com/codefordc/project-management/tree/master/python

```
Traceback (most recent call last):
  File "python/scripts/load_data.py", line 14, in <module>
    from housinginsights.ingestion.DataReader import DataReader
ImportError: No module named 'housinginsights'
```

In that part of the code, when you run it locally I have a line that appends the `housing-insights/python` folder to the path so that the package located at  `housing-insights\python\housinginsights` can be imported. I tried debugging a bit but couldn't get it working. I'm new to Docker so I'm pretty slow at debugging this stuff. 

[...]

PR #245, opened 2017-05-08
Add sandbox Dockerfile for running python code

jasonrhaas:
## What's new

This fixes issue #183.  I added a new Dockerfile that has a pre-built environment for running all of our Python code.  Now we can run the python code inside this container without installing Miniconda locally.  This should make it easier to get our code up and running quickly.

## Testing it out

In the `housing-insights` directory, 
- In your **secrets.json** file, change the `docker_database` string to work inside the docker container. 
```
    "docker_database": {
        "_comments": "You may need to change the IP address and/or port for your machine.",
        "connect_str": "postgresql://codefordc:codefordc@postgres:5432/housinginsights_docker"
    },
```

- `docker-compose up -d`.  This will bring up all the containers including the `sandbox` which is the miniconda3 python container.
- `docker-compose exec sandbox bash` to get into the docker container
- `cd /code/python/scripts` to get into the script directory to load the data
- `python load_data.py docker rebuild sample` to load in the sample data

**NOTE**:  The data loading does not work at the moment due to some import errors.  This error is the same in the local environment and docker environment so it should not be an issue with the container.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,3 +30,10 @@ services:
         ports:
             - "5432:5432"
         restart: "on-failure:3"
+
+    sandbox:
+        build:
+            context: python
+        volumes:
+            - .:/code
+        command: tail -f /dev/null
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -1,3 +1,17 @@
-FROM python:3.5-onbuild
-COPY . /
-LABEL maintainer "NealHumphrey"
+FROM continuumio/miniconda3
+
+MAINTAINER Jason Haas <jasonrhaas@gmail.com>
+
+# Set up code directory
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+# Copy over files
+COPY environment.yml .
+COPY requirements.txt .
+
+# Install dependencies
+RUN conda env create
+
+# Activate environment
+RUN echo "export PATH=/opt/conda/envs/housing-insights/bin:$PATH" >> /root/.bashrc


docs/resources/onboarding/docker.md
```diff
@@ -54,6 +64,14 @@ This will print the IP address of your local Docker instance. Open up `/python/h
...
+
+```
+    "docker_database": {
+        "connect_str": "postgresql://codefordc:codefordc@localhost:5432/housinginsights_docker"
+    },
+```
+
 **Troubleshooting note**: If you already have Postgres installed on your machine, it may be using the 5432 port. If you get an error message saying something related to a port in use, you'll need to change the port both here AND in the docker-compose.yml file. 
 
 ### 1b) Day-to-day use
```
NealHumphrey: I think the line 90 below (new numbering) should change to `docker-compose up -d` to allow access to the docker command line, right?

docs/resources/onboarding/docker.md
```diff
@@ -87,6 +105,30 @@ If you've got Docker running, you should be able to see your database in the web
 
 ### 2a) One time setup
 
+Here again there is the option to run python inside the `docker-compose` environment or using your local python setup.  The easiest method is to run using the first option below, but the others will work as well.
+
+**Full Docker Compose (docker Python)**
+
+1)  `docker-compose ps` to see the containers that are running.  One of them should be **housinginsights_sandbox_1**.  This is the container that is running the miniconda3 python environment.  There is no need to activate or deactivate an virtualenv since it is already sourced.
```
NealHumphrey: Can you add a note clarifying that this is the same command line (or Docker Quickstart Terminal when using Toolbox) as the one that you ran docker-compose up on? I expect there will be confusion since the local option (which is what people are used to using) used two separate terminal windows.

NealHumphrey:
@jasonrhaas Thanks for this! I'm going to have to wait until tomorrow to test this locally but should get this in tomorrow. Documentation looks overall great, there's just the one note I had about clarifying where to type the commands (single vs. two command prompt windows). Can you verify my interpretation was correct and edit if so?

docs/resources/onboarding/docker.md
xee5ch: If you run `docker-compose up -d` I was under the impression it runs it all in the background; the `-d` is for detached after all.

xee5ch:
Again, I am not sure if I am doing this wrong, but this is exactly the same issue I ran into previously when trying to build a docker-compose worfklow using `build` with `Dockerfile` directives with manually copied files (in your case environment.yml and requirements.txt are in properly in `/usr/src/app` as desired) and `volume` directives (in your case `/code` is empty: you either get the one or the other, but not both.

```
[me@craptop housing-insights]$ docker-compose down
[me@craptop housing-insights]$ docker rmi SANDBOXHASH # just to see I am not crazy
[me@craptop housing-insights]$ docker-compose build
[me@craptop housing-insights]$ docker-compose run sandbox bash
Creating network "housinginsights_default" with the default driver
root@ce721b10be2e:/usr/src/app# 
root@ce721b10be2e:/usr/src/app# pwd
/usr/src/app
root@ce721b10be2e:/usr/src/app# ls
environment.yml  requirements.txt
root@ce721b10be2e:/usr/src/app# ls -la /code/
total 4
drwxr-xr-x  3 root root   60 May  9 01:50 .
drwxr-xr-x 47 root root 4096 May  9 02:14 ..
drwxr-xr-x  2 root root   40 May  9 01:50 docs
root@ce721b10be2e:/usr/src/app# ls -la /code/docs/
total 0
drwxr-xr-x 2 root root 40 May  9 01:50 .
drwxr-xr-x 3 root root 60 May  9 01:50 ..
root@ce721b10be2e:/usr/src/app# 
```

Correct me if I am wrong, but we want the python code added to the container, no? I get the feeling I must be doing something wrong, but I found lots of warnings about workdir and volume overlay not going as planned, it's just not clear if it is with docker-machine or not. I used machine on purpose to make sure I get the same environment as the Mac and Windows people as well, so as I discussed with Neal on Sunday, I feel I am beating a dead horse or I am misreading docker-compose/machine docs wrong.

docs/resources/onboarding/docker.md
jasonrhaas: Oh yea, I always run `docker-compose up -d`, and do `docker-compose logs <name>` to access the logs if I need to.  If things are running properly, I usually don't care about seeing the terminal output.

I can change the command all to `docker-compose up -d` if you are good with that.

docs/resources/onboarding/docker.md
jasonrhaas: I think we should just change everything to `docker-compose up -d` and then we don't have to worry about the confusion.

jasonrhaas:
@xee5ch Are you running all the latest versions of Docker and docker-compose?  Are you running on Windows, Mac, or Linux?  I am sure that the above implementation is working properly on Mac and Linux, but I have not tried it on Windows.  Maybe the volume mounts work a little differently with Windows, or maybe its a system config thing?

As far as copying the python code over to the image, we shouldn't do this unless we are running production code inside the container.  This container is only used for development purposes -- copying code over makes it difficult to do development on, because anytime you make a change to the repo code you need to rebuild the container, which takes time and is easy to forget.  And you may find your self in a situation where you think you are running the latest code inside the container, but really its old code, and stuff is broken and then you waste time trying to figure out why.

xee5ch:
You're all good. It turns out the Linux version of docker-machine, of using the underlying VBox defaults even if guest support is configured (I did so) fails at filesystem sharing. That was the root cause all along.

I stayed up late and went back to normal Docker on Linux and solved networking issues I had. I was trying to stay on the same page as everyone else. Sorry for wasting time.

TL;DR docker-machine on Linux cannot share /home like it does /Users on on OSX or Windows in many cases so its looking for the files in the docker-machine VM. This is a known limitation and those like me should know better.

I tested your stuff and it works well.

jasonrhaas:
Ahhh yea the old guest support thing with Virtual Box.  As soon as native support for Docker was available for Mac, I switched to that and never dealt with VM's again.  Same thing with Linux, just run Docker natively, and it makes things a lot easier.  While Docker doesn't solve everything, it has _almost_ killed the need to test in VM's, which is 💯 in my book.

docs/resources/onboarding/docker.md
jasonrhaas: @NealHumphrey are you good with the above comment?

NealHumphrey:
Thanks @jasonrhaas ! Tested and works great. I'm making a couple more documentation edits and will tell everyone about the new option in the update email today.

PR #253, opened 2017-05-09
WIP: javascript refactor — data overlays working, needs to check that layer does not exist …

ostermanj:
…before adding it.

This continues work on the refactor, getting close to the existing functionality of the alpha tool.

This PR has the data overlays working (crime and building permits), with some details to work out.


ostermanj:
Latest commits separates the code for different views into different files. housing-insights.js houses the controller, pubsub, and setState/getState functions; map-view.js handles the map-view; side-bar-view.js the sidebar, etc.  Should be easier for collaboration.

The refactor now has all the functionality of the alpha-version tool except the legend is missing. I din't want to hard code it in, but I haven't yet figured out how to create mapBox legend that pulls dynamically from a layer with data-driven paint styles. all the mapBox examples seem to be hard-coded, which isn't great. One solution may be to create the legend first using d3 and use that process to pass the relevant properties to mapBox for it to render the layer.

ostermanj:
Clarification: this has the functionality for the map-view, not for the building view. Clicking on the link in the popup publishes a state change that should be the trigger to swicth to the building view, but that's not it yet.

NealHumphrey:
@ostermanj I'm going to dig into the code again later today, but overall I'm really excited about this approach. 

I did a quick test of pulling this version up in the browser, Pie charts and chloropleth are working great, but the project points aren't showing up. Is this expected or not? I'm trying to figure out what exactly we can have in the codebase for tomorrow's hackathon.

ostermanj:
Hmm. That's unexpected. They're still working in my local 212-jo-sandbox-option branch of javascript/refactoring_sandboxes/option2/. They are not working for me  in Paul's  docs/tool/ version.

NealHumphrey:
@ostermanj hmm. I'm at this conference so can't really dig into the code to troubleshoot. If it's working on your end, the best path is probably just resolve the issues in @ptgott 's merged version that is moved into docs and try to get everything working there and see if the problem goes away. Sounds like both of you have some time to look at it today so maybe coordinate timing to decide if one or both of you should take a crack at it - it does seem like it's probably simple import file renaming

NealHumphrey:
@ostermanj  my mistake! I was using the wrong version. Your version is working all correctly!

Issue #198, opened 2017-04-13
Tax assessment data via opendata API


NealHumphrey:
Opendata.dc.gov provides [tax assessment data](http://opendata.dc.gov/datasets/integrated-tax-system-public-extract-facts) of all properties in DC via an API.

- [ ] Create a TaxApiConn class based on the MarApiConn class currently in the code. Use this to add raw data files to /data/raw/tax_assessment/opendata/YYYYMMDD when the python/cmd/data.py script is run with the appropriate arguments. Write a demo command to run at the command line to create this file with the appropriate datestamp.
(pull request after step 1; someone else can do part 2 if desired)
- [ ] Add this file to the manifest.csv using the instructions on [adding a new dataset](http://housinginsights.org/resources/adding-data-sources.html).

Future: automatically add new files to the pipeline whenever this update script is run, but we wait to do this until we have created an appropriate structure for all API scripts.

NealHumphrey:
Note on integrating this data from open.data@dc.gov:

EXTRACTDATE field says when this was last updated, looks to be roughly monthly.

NealHumphrey:
Hey @eng1nerd have you been able to do any work on this issue? Let me know where things stand - and if you're busy with the new job we can also see if someone's able to take it over!

eng1nerd:
@NealHumphrey I'm sorry for the delay. I will try to finish it by tomorrow 6pm. If something will not be right or it will take me much longer, then some other person should probably step in.

NealHumphrey:
Latest status:

- @eng1nerd 's code has been merged into the codefordc repository under the branch name `198-add-tax-data`; should pick up code from there. 
- Current TaxApiConn class uses the 'GeoService` api. Instead we should use the 'GeoJSON' api url, but swap it out for .csv as noted on the opendata documentation (bottom of page): http://opendata.dc.gov/pages/using-apis . This will resolve the issue that the number of rows is limited in the geojson api (and also make it easier for us to parse since it'll be in the format we want anyways).

PR #256, opened 2017-05-10
Added Tax assessment data via opendata API #198

eng1nerd:
TaxApiConn class created.
`python data.py tax get_tad` is command to create file with the current datestamp.

### DIFF
--- a/python/housinginsights/sources/models/tax.py
+++ b/python/housinginsights/sources/models/tax.py
@@ -1,0 +1,27 @@
+'''
+Model for tax api
+'''
+from datetime import datetime
+
+FIELDS = ['OBJECTID', 'SSL', 'ASSESSOR_NAME', 'LAND_USE_CODE', 'LAND_USE_DESCRIPTION',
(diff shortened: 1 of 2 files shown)


NealHumphrey:
@eng1nerd Thanks for this! It's a good start, but it looks like opendata.dc.gov is deciding not to play nicely with us. The csv file created only had 1,000 rows - I looked into this and it appears this is an ESRI default from their geojson api: https://geonet.esri.com/thread/67749

It turns out that the csv files we've been manually downloading from opendata.dc.gov can actually be queried directly from their API, which is the better way to get a bulk download without limits. They note that it's just a matter of swapping .geojson for .csv in the geojson api endpoint:
http://opendata.dc.gov/pages/using-apis (bottom of page)

I merged your code into the 198-add-tax-data branch instead of into dev. That way we have it in the codefordc repo but can pick up resolving these issues.

NealHumphrey:
@eng1nerd You mentioned you were busy with starting this new job - do you want to try changing this to the .csv approach or would you rather I put this back out to the team?

PR #259, opened 2017-05-12
ready for review: Made wmata_distcalc.py OO and made it ingest directly from the docker…

jgordo04:
Made wmata_distcalc.py OO and made it ingest directly from the docker database instead of a file


python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -0,0 +1,258 @@
...
+class railStation():
+
+    def __init__(self,wmata_api_key,mapbox_api_key):
+        self.wmata_api_key = wmata_api_key
+        self.mapbox_api_key = {'access_token':mapbox_api_key}
+
+    def getWmataHeaders(self):
+        return { 'api_key': self.wmata_api_key}
+
+    def writeRailInfo(self, infoCsvWriter):
```
NealHumphrey: If you inherit from the BaseApiConn, you get access to the result_to_csv method. It'd be worth seeing how hard it would be to refactor this to use the same approach, mostly for consistency/maintainability.

python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -0,0 +1,258 @@
...
+            lines = ""
+            for route in stop['Routes']:
+                lines = '{}:{}'.format(lines, route)
+            lines = lines[1:] #take off the first :
+
+            infoCsvWriter.writerow((stop['StopID'], 'bus', stop['Name'], stop['Lat'],stop['Lon'], lines)) #, stop_id_or_station_code))
+    
+
+def main(secretsFileName, distOutputFileName,infoOutputFileName):
+    """Writes two csvs: 1 for general bus/rail info, 1 with distances to wmata for projects
```
NealHumphrey: Ok, so this is the big one - basically everything in main() needs to move into the classes above. This way everything we need to get the data is contained here, and we can just say 'for each ApiConn object, get our data' and all the apiConn classes act the same way. A few other comments above align with this too. 

Goal (pseudo code):
```
for api_source in (TaxApiConn,HousingApiConn,CrimeApiConn,WmataInfoApiConn,WmataDistApiConn):
     api_source.get_data
```

python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -0,0 +1,258 @@
...
+
+   Parameters:
+   secretsFileName - json file name that contains various api keys
+   csvInputFileName - csv file with project information
+   distOutputFileName - csv file to output to for calculated metro distances for each project
+   infoOutputFileName - cvs file for general bus & rail info for each wmata station
+   """
+
+    try:
+        conn = psycopg2.connect("dbname='housinginsights_docker' user='codefordc' host='192.168.99.100' password='codefordc'")
```
NealHumphrey: Oh, also - should use the sqlAlchemy connection instead of psycopg2, just since that's what we're doing elsewhere. That's what that funciton returns.

NealHumphrey:
Hey @jgordo04 I renamed this to WIP to help me keep track of the status of which pull requests need review and which have changes needed. Let me know when you're ready for me to review again!

jgordo04:
Hey @NealHumphrey . I pushed an updated version to my dev repo (and I think it automatically updates the pull request?). If not, let me know and I'll do another pull request. Several things I'm still working on.
1. using results_to_csv from base class
2. adding optional parameters to choose which database to use (still working on that)

python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -0,0 +1,234 @@
...
+        numrow = 0
+
+        for row in rows: #reader:
+            radius = self.getMeters(0.5)
+            self.setProjectInfo(row,columns)
+
+            numrow = numrow+1
+
+            print("Processing project {} of 400ish".format(numrow))
+
```
NealHumphrey: could say how many it actually is - wasn't possible in the old code because the DataReader used an iterator, but now getting it from the database we know there's `len(rows)`

python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -0,0 +1,234 @@
...
+        print("Will write WMATA_DIST table to {}".format(self.distOutputFileName))
+        print("Will write WMATA_INFO table to {}".format(self.infoOutputFileName))
+
+        #pull API keys
+        self.api_keys = json.loads(open(secretsFileName).read())
+        self.wmata_api_key = self.api_keys['wmata']['api_key']
+        self.mapbox_api_key = {'access_token':self.api_keys['mapbox']['public-token']}
+
+        #write out the wmata info csv
```
NealHumphrey: Should probably move these csv initialization lines into get_data. Currently it's not a concern because we instantiate the object and then use it immediately, but there might be situations where we want to call get_data later or in a different context.

python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -0,0 +1,234 @@
...
+
+        for station in railStations:
+            #delimit list of lines with colon
+            lines = station["LineCode1"] #there is always at least one station
+            for line_code in ["LineCode2", "LineCode3", "LineCode4"]:
+                if station[line_code] != None:
+                    lines += ":" + station[line_code]
+            infoCsvWriter.writerow((station['Code'], 'rail',station['Name'],str(station['Lat']), str(station['Lon']),lines))
+
+        return railStations
```
NealHumphrey: You should make writeBusInfo and writeRailInfo behave in the same way - either they both return their json stations response, or neither does.

python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -0,0 +1,234 @@
...
+                  'Radius':str(radiusinmeters)}
+        response = requests.get('https://api.wmata.com/Bus.svc/json/jStops', params=params, headers=wmata_headers)
+        data = response.json()
+
+        for stop in data['Stops']:
+            walkDist = self.getWalkingDistance(lat, lon, str(stop['Lat']), str(stop['Lon']))
+            if walkDist <= radiusinmeters: #within 0.5 miles walking
+                distCsvWriter.writerow([Nlihc_id, 'bus', stop['StopID'], "{0:.2f}".format(self.miles)])
+
+    def writeRailInfo(self, infoCsvWriter):
```
NealHumphrey: Maybe we can rename this to get_all_rail_stations? That appears to be what it's doing, right?

python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -0,0 +1,234 @@
...
+            lines = station["LineCode1"] #there is always at least one station
+            for line_code in ["LineCode2", "LineCode3", "LineCode4"]:
+                if station[line_code] != None:
+                    lines += ":" + station[line_code]
+            infoCsvWriter.writerow((station['Code'], 'rail',station['Name'],str(station['Lat']), str(station['Lon']),lines))
+
+        return railStations
+
+    def writeBusInfo(self, infoCsvWriter):
+        """Writes all bus station data to a given CSV writer.
```
NealHumphrey: Same thing - get_all_rail_stations?

python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -0,0 +1,234 @@
...
+        self.infoHeaders = ('code_or_id','type','name','lat','lon','lines','stop_id_or_station_code')
+        self.infoCsvWriter.writerow(('code_or_id','type','name','lat','lon','lines','stop_id_or_station_code'))
+
+        #write out hte dist.csv file
+        self.distOutputFile = open(self.distOutputFileName, 'wt')
+        self.distCsvWriter = csv.writer(self.distOutputFile)
+        self.distHeader = ('Nlihc_id','type','stop_id_or_station_code','dist_in_miles')
+        self.distCsvWriter.writerow(('Nlihc_id','type','stop_id_or_station_code','dist_in_miles'))
+        
+        #Creat railStation object
```
NealHumphrey: We should move this out of init and have the user call it directly instead. That way they can choose whether or not to update the wmata info data.

python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -0,0 +1,234 @@
...
+        self.railStations = self.writeRailInfo(self.infoCsvWriter)
+        self.writeBusInfo(self.infoCsvWriter)
+
+    def getData(self):
+        try:
```
NealHumphrey: Should we rename this to get_station_walking_distances? See below for other renamings.

Then, we can add a new function called get_data that just does this:
```
self.get_rail_stations
self.get_bus_stations
self.get_station_walking_distances
```

jgordo04: @NealHumphrey ... not sure how you would want this to work. get_data calls _find_rail_stations and _find_bus_stations for each row in the database. How do you propose taking those out of the loop? Or just looping within each method? That would take twice as long, since it would have to loop through twice, no?

python/housinginsights/sources/wmata_distcalc.py
jgordo04: @NealHumphrey ... you mean the distHeader definition (which for some reason I wrote out a second time in the writerow line...)

python/housinginsights/sources/wmata_distcalc.py
jgordo04: made these both _get_all_[rail/bus]_stations

python/housinginsights/sources/wmata_distcalc.py
jgordo04: they both return json now. I want to also pass the bus json to _get_all_bus_stations so that I don't have to do the query a second time, but it has slightly different parameters, so I'm just going to make sure it doesn't make a huge difference first (on my To Do list)

python/housinginsights/sources/wmata_distcalc.py
jgordo04: Renamed

python/housinginsights/sources/wmata_distcalc.py
jgordo04: Done

python/housinginsights/sources/wmata_distcalc.py
jgordo04: Done

python/housinginsights/sources/wmata_distcalc.py
NealHumphrey: I was talking about lines 52 and 53 -self.railStations and self.writeBusInfo. Since each of these methods actually query the API and then write the information to CSV's, we shouldn't have them run during object initialization; instead, they should be user driven and callable if desired.

python/housinginsights/sources/wmata_distcalc.py
NealHumphrey: @jgordo04 this is linked to the other comment you asked about, moving the writeRailInfo and writeBusInfo out of the __init__. I was saying that get_rail_stations == writeRailInfo (hopefully get_rail_stations is clearer name? go with whatever name you prefer), and get_bus_stations == writeBusInfo. 

get_station_walking_distances == getData, which would pretty much be unchanged. As you point out it still needs to use the findRailStations which as you point has to be called for each row or the project table. Continuing the spirit of explicit naming, findRailStations might be renamed to find_nearby_rail_stations or something.

NealHumphrey:
@jgordo04 I clarified a couple points you asked about, hopefully that shows what I meant! Let me know when this is ready for me to test locally! Looks like it's almost there.

jgordo04:
@NealHumphrey Split get_data up into get_dist and get_info and get_data just calls either of those. ALso made it so that you can spit data out onto console instead of into the csv files (though I don't know if you actually wanted that). Hopefully, this pulls everything together!

python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -0,0 +1,295 @@
...
+
+        wmata_headers = self._get_wmata_headers()
+        railResponse = requests.get("https://api.wmata.com/Rail.svc/json/jStations", headers=wmata_headers)
+        self.railStations = railResponse.json()['Stations']
+
+        for row in rows: #reader:
+            radius = self._get_meters(0.5)
+            self._set_project_info(row,columns)
+
+            numrow = numrow+1
```
NealHumphrey: The cleaner Python way of doing this is to use enumerate on the for loop:

`for idx, row in enumerate(rows):`

Then you don't have to initialize numrow or remember to add 1 to it (easy to forget).

python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -0,0 +1,295 @@
...
+        lat = self.lat
+        lon = self.lon
+        Nlihc_id = self.nlihcid
+
+        for station in railStations:
+            walkDist = self._get_walking_distance(lat, lon, str(station['Lat']), str(station['Lon']))
+
+            if walkDist <=radiusinmeters:
+                #railData.append([Nlihc_id, 'rail', station['Code'], "{0:.2f}".format(self.miles)])
+                self.distOutput.append([Nlihc_id, 'rail', station['Code'], "{0:.2f}".format(self.miles)])
```
NealHumphrey: Oops, this should write out `walkDist` not self.miles (fixed)

NealHumphrey:
Closing this pull request because I have merged teh code into #286 which combines all the latest apiconn classes into one big PR with some cleaned up method signatures and common functionality. When that PR gets approved the commits will show up as coming from this branch (so you'll still show up in the contributors stats)

python/housinginsights/sources/wmata_distcalc.py
NealHumphrey: note, in my fix I changed it to walkDistMiles and applied the conversion

Issue #237, opened 2017-05-04
JUNE: BROWSE Master Issue


NealHumphrey:
For our June 13th launch, the Browse team is in charge of both populating our database with the bulk of our data sources, as well as building out the front end navigation to toggle between viewing multiple data sources. These are the components that need to be completed in time for our June 13th launch:

Data quality
- [x] Geocode all data sources with a consistent format for the city zones (ward, zipcode, etc.)
- [x] Add nlihc_id to any data source that could match a record from the project table
- [x] Coordinate with the Filter team for the creation and use of a 'Neighborhood Facts' table (related, #236)

Ingestion infrastructure
- [x] Add a system to update all data from APIs with one function call. 
- [x] Design system for connecting downloaded API flat files to our current ingestion method
- [x] Encapsulate load_data.py into a class or function that can be callable from another script #257 
- [x] Make sure that data is appended or overwritten as needed. #132 
- [ ] Utility for joining new meta.json #196 

Data Sources
- [x] Make a complete plan for data sources ingestion #188
- [ ] Add all additional affordable housing records from DC gov sources
    - [ ] DMPED's database available from opendata.dc.gov #194 
    - [ ] DHCD's pipeline database #136 
    Additional sources of affordable housing projects (as [described here](http://housinginsights.org/data/affordable-housing-inventory.html) can wait until the V2 launch. 
- [x] WMATA data from API #200 
- [x] Tax Assessment data #198 
- [x] Crime data #187
- [x] All 5 PresCat tables
- [x] Census tables #152 
- [x] All housing units (non-subsidized) #140 
- [ ] Zillow market rent data #53 and #15
- [ ] PUD applications #50 (should make new, clean ticket)
- [ ] Building Permits #16 (should make new, clean ticket)

Browse view
- [x] Toggle between data sets (probably similar sidebar list that can slide in/out [like this one](http://datatools.dcactionforchildren.org/) or similarly [this one](https://ccusa.github.io/Disaster_Vulnerability_Map/))

ptgott:
@NealHumphrey @ostermanj @jkwening @dogtoothdigital 
Today I've been working on changing around rich-map.js to accommodate a more extensible range of data sets. It's all really messy so far, but I should have a PR ready by Tuesday's meeting. The main goals are:
- Query the API as menu options are selected, rather than loading all data sets at once at the start.
- Use a JSON file to populate the menu of data sources and zones. This way people adding datasets to the API can make them available to the front-end without getting into the JS.
[...]

PR #261, opened 2017-05-14
237 decouple datasets from browse

ptgott:
I've summarized this PR in issue #237.

I unthinkingly merged the dev branch into my topic branch before realizing that I hadn't squashed the commits made on 6 - 7 May. Sorry for the messy commit history!


NealHumphrey:
@ptgott thanks for this! 

We'll need to figure how to merge this with @ostermanj 's changes. Considerations:

- Before the Saturday Hackathon, I want to make sure we can re-open the main /tool folder for editing by all, with our refactored code mostly stabilized. 
- It looks like John's putting all his stuff into the option2 folder (treating that folder like a clean slate version of the /tool folder)
- This one works directly in the /tool folder 
- We have the outstanding JS tests PR that Paul made - should bring that in too
- Before Saturday, I think we should go ahead and move the /javascript/tool folder into the /docs/tool folder and remove the /javascript folder, unless either of you have opposition. That way a) we can do things like build in appropriate links to the data documentation pages as well as b) note have to copy-paste to deploy. Downsides are c) everyone will need Jekyll (hopefully Docker will be less painful) and d) we'll need to be much more diligent about when pull requests will impact the live site. 

Will both of you be there tomorrow night? I'd suggest that we take Johns code, move it into a Jekyll structure, and then see if Paul can manually bring in these changes to those files. Thoughts?

ostermanj:
Hey all. I should be able to move my sandbox over to the docs/ folder
before tomorrow's work session. I may be quite a bit late tomorrow but hope
to be able to make it by 8. It would be good go over everything with the
front-end team.

NealHumphrey:
@ostermanj Do you think you'll do any changes besides moving the folder? Just want to know what changes to expect from the current code if I have time to dive in and learn what you've made tomorrow.

ostermanj:
Probably no big changes. I'd like to make use of the manifest and work on
the legend functionality.

NealHumphrey:
Do you mean make use of the [meta.json](https://github.com/codefordc/housing-insights/blob/dev/python/scripts/meta.json)? 

I'm working on that right for the data documentation pages. Hope to have it up tonight. That would show the Jekyll methodology of using it, shouldn't be hard to duplicate in JS.

ostermanj:
Cool. Yeah that's what I meant.

ptgott:
I'm okay with moving the /javascript content into /docs!

As for merging this PR with the refactor, I can jump on this before Saturday and hopefully incorporate the key features. Let's set some priorities tomorrow for which parts of this PR to bring in.

ostermanj:
Great stuff! We'll definitely want to bring in the function to check for what type of data combinations are available for use. I also really like the toDom functions for each view. The biggest difference in our approaches is that @ptgott is using constructors to create each map layer, which may be more efficient coding.

NealHumphrey:
@ostermanj FYI - I couldn't get the meta.json to work with Jekyll's templating engine - basically meta.json was in a slightly wrong format (key-value object instead of array) and I didn't want to alter it and maintain two copies. We'll want to use Javascript to get the meta.json from the API and put it on the page instead - so however you use it we'll want to also apply to the documentation pages. I haven't started the JS for that, just tried the Jekyll approach.

ostermanj:
I was wondering how to access the meta.json since it's under the python/
folder, not docs/. So it'll be available by API? I wonder if would query it
for a subset as needed or fetch the whole thing at the beginning?

NealHumphrey:
For now, plan on getting the whole thing - it makes it easier on the API side since we still might do some refactoring on the back end.

PR #266, opened 2017-05-17
Meta.json in the API

NealHumphrey:
- Added function to load meta.json into the database and call that from load_data.py
- Added api endpoint to serve meta to the front end
- Added optional command line argument to get API data from docker for testing purposes

As of 5/17 AM this has not yet been deployed to live site.

If someone's able to review this PR let me know! @ohaaga or @jasonrhaas or @jkwening

### DIFF
--- a/python/api/application.py
+++ b/python/api/application.py
@@ -16,6 +16,7 @@ import logging
 from flask_cors import CORS, cross_origin
 
 import math
+import sys
 
 #Different json output methods.
 # Currently looks like best pick is jsonify, but with the simplejson package pip-installed so that
@@ -56,9 +57,15 @@ application.json_encoder = CustomJSONEncoder
 #Allow cross-origin requests. TODO should eventually lock down the permissions on this a bit more strictly, though only allowing GET requests is a good start.
 CORS(application, resources={r"/api/*": {"origins": "*"}}, methods=['GET'])
 
+#Allow us to test locally if desired
+if 'docker' in sys.argv:
+    database_choice = 'docker_database'
+else:
+    database_choice = 'remote_database'
+
 with open('secrets.json') as f:
     secrets = json.load(f)
-    connect_str = secrets['remote_database']['connect_str']
+    connect_str = secrets[database_choice]['connect_str']
 
 
 #Should create a new connection each time a separate query is needed so that API can recover from bad queries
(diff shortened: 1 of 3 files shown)


jkwening:
@NealHumphrey @ohaaga @jasonrhaas I can review it later this afternoon - it will give me an opportunity to better understand the code base and the api

NealHumphrey:
thanks @jkwening ! 

To test:
1) Rebuild your docker database `load_data.py rebuild docker`. Or you can skip rebuilding and it will just load new tables only i.e. `load_data.py docker` (compares manifest.csv to manifest table and only loads new files). This will add a new table to the database (meta) with one column (also named meta) and one row (the whole meta.json stuffed into one cell)
2) Serve up the api from the /python/api folder, using the docker database instead of remote one `python application.py docker`
3) Visit the url for the meta endpoint `192.168.99.100:5000/api/meta`

NealHumphrey:
Hey @jkwening have you had a chance to review yet? I'm hoping to deploy before tomorrow so we can have this available to front end people during the hackathon tomorrow.

python/api/application.py
```diff
@@ -56,9 +57,15 @@ def default(self, obj):
 #Allow cross-origin requests. TODO should eventually lock down the permissions on this a bit more strictly, though only allowing GET requests is a good start.
 CORS(application, resources={r"/api/*": {"origins": "*"}}, methods=['GET'])
 
+#Allow us to test locally if desired
+if 'docker' in sys.argv:
+    database_choice = 'docker_database'
+else:
+    database_choice = 'remote_database'
+
 with open('secrets.json') as f:
```
jkwening: I ran into an issue with this as a relative path - I think this needs to be turned into a relative path in reference to the application.py file path. Unless I'm missing something, currently you'd have to be in `python/housinginsights` path and then run `python ../api/application.py docker` for it run correctly.

NealHumphrey: @jkwening Good point. You should actually just copy-paste your secrets.json over into the api folder (duplicated). It doesn't show up in git because secrets is excluded. 

Double maintenance of secrets will be resolved when we refactor this in the future - eventually the API server will need access to the whole python folder so that it can run data collection jobs.

jkwening: @NealHumphrey ahhh...gotcha. Then this is good to go. Thanks.

PR #269, opened 2017-05-18
WIP: Dynamic zone type disabling

ptgott:
As discussed in issues #212 and #237. It turned out that I didn't really have to introduce new features (I just understood the refactor less than I thought!). The bulk of this is rearranging the code to grab data from meta.json, rather than from the 'manifest' object. This meant changing controller.getData and the conventions for dataRequest (you build the URL for the d3.json call prior to calling getData). The code populates the array of initial overlay layers dynamically from meta.json.

I'm incorporating the refactor in this pull request, rather than in @ostermanj 's repo, because to do the latter would be at the limits of my git know-how. I've moved the refactor into docs/tool/js. PR #261 is in the original javascript/tool/js directory. I have not needed to copy anything from it.

I've copied meta.json into docs/tool/data to expose it to the Jekyll server. This includes new keys related to the api in crime, project and building_permits. We can indicate which datasets are available to which api endpoints, and the map view will populate the datasets menu automatically from meta.json.

**There are still loads of runtime errors.** I can get to fixing these tomorrow if you'd like. At the moment this PR isn't usable. 

**Other things to do later**
1. Gray out any menu options not available for a given layer. This should be straightforward, as the code populates the array of overlays with objects that include the available aggregation zones for a given dataset.

2. I'm not sure if the building view will work with the new version of the refactor.


NealHumphrey:
Hey @ptgott - thanks for working on this! Yes, please do see if you can get these exceptions handled - it looks like it's probably mostly file structure issues related to moving the files over. Did you delete the contents of /docs/tool and paste in John's version? Or did you add files one by one? 

When do you think you'll get to this? Do you think it can be done today? I can some work on it this afternoon if needed - but of course don't want to duplicate any work you do. 

I changed the name to WIP, let me know and change the name back when it's ready to try merging in.

ostermanj:
@ptgott @NealHumphrey I can review today. I think some of the errors are from field names changing on the API side.

ptgott:
@nealhumphrey @ostermanj
I pasted all of the refactor docs into the corresponding /docs/tool folder.

I'll do some more work on debugging this morning and let you know how things shape up!

ostermanj:
@ptgott: one of the errors is resolved by changing the `neighborhood` key
in the zoneMapping object (side-bar-view-.js) to `neighborhood_cluster`. I
imagine others are due to similar difference between what the API is now
returning and what was in the static files.

On the overlays, in line 99 of map-view.js, there shouldn't be quotes in
the bracket notation of `data.overlay`, right? Fixing that gets me closer
to having the overlays working, but still with errors in addToGeoJSON.

I haven't looked into why the project circles aren't showing.

So if you can debug this morning I can pick up some this afternoon.

ptgott:
@ostermanj Thanks for looking at this! I'll incorporate those edits.

I've got the project circles to show simply by changing the name of the dataRequest in sideBar.init to 'raw_project' from 'raw'. A lot of these errors also look like issues with getting names of things to match up after I rearranged things, as well as oversights like the quotes in the brackets.

ptgott:
@ostermanj @NealHumphrey 
Right, so I've got the overlays and the initial project dots to display again. The overlays issue was down to the '<table>_all_<grouping>' syntax from the refactor no longer applying to the new dataCollection objects fetched with the tweaked getData code. 

I'm going to spend the next hour seeing if I can get unavailable dataset options to go gray depending on the zone and vice versa, then pass this on to @ostermanj.

Some more things to do include copying the legend over from the old code, as well as getting links to the building page to work (I'm not sure how broken this will be since the refactor and this PR!).

ostermanj:
ok - I can get the legend in there, hopefully dynamically created but probably hard-coded. I still don't see a mapBox-native way to do it in a non-static way (strange!) but may be able to think of something.

And I should be able to the building view to come up, at least in the same way it was working before as a separate URL.

@NealHumphrey : when should we stop making changes for you to prepare for tomorrow?

ptgott:
Great, thanks @ostermanj! [This](https://github.com/codefordc/housing-insights/pull/269/commits/38db236bd2aa8f1692b0e977c020a74733ca22f3) commit extracts menu options for layers and overlays into constructors so we can toggle between unavailable and available. It's pretty buggy, but most of it's there.

NealHumphrey:
@ostermanj I'll be at home at 3pm and it would be great to have a mergeable
version then if we can. I'll be digging into the current version on the
train ride so keep pushing commits as you go. For tomorrow the big thing is
to make sure the structure is in place so that the filter functions we'll
be adding can tap into the StateModule appropriately.

ostermanj:
@NealHumphrey @ptgott I can add to this PR directly, so I made another,  #270 . Description over there.

NealHumphrey:
Hey @ptgott with @ostermanj 's PR #270 almost ready to merge in, I renamed this pull request to reflect the additional functionality left on this branch that John couldn't incorporate yet (extract menu options). Feel free to close this PR and open a new one with that code if you prefer!

PR #270, opened 2017-05-19
Ostermanj 212 browse

ostermanj:
@NealHumphrey @ptgott I couldn't change @ptgott 's PR directly and so had to create a new one.

This puts the legend back in (hard-coded 👎   ) and reestablishes the linking to the building.html after clicking through on a building. There's code in there that builds out how this could be done in the future without loading a new page, but the building view needs a lot of work in response to the refactor to make that work. That code is commented out now in deference to the old way of loading building.html with a query parameter. You can reverse the comments to see how it could work. Right now building.html links to an old version of housing-insight.js (explicitly housing-insight-OLD.js) so that it works as before.

I've removes the Zillow option since it throws errors in relation to the donut charts (donut data is not available in zillow zones). I've also removed @ptgott 's very good start on graying out unavailable options because I didn't have time to look at the buggy-ness. We can bring it back in later. But also, I wonder if people are more likely to select zones first, then data sets. I.e.: should the overlay menu change according to which zone is selected, rather than vice versa? Not sure.


NealHumphrey:
Awesome! I'm going to do a bit of testing and see if I can break anything critical or see any major impediments to making this the official version and then will merge in. 

On your building view approach - sounds good, I think it is fine to kick this can down the road for now given that we did not have enough people to establish a building view team for V1. 

On your question of which order people will select zone vs. data, my guess is the typical pattern will be:
1) select data type (we should default to a zone type)
2) select different zone type (we should persist the data source that was last selected)
3) select second data type (we should persist the zone type that was last selected)
3b) select second data type that does not have aggregation available at the last selected zone type (should switch zone type to default zone type for that data set).

NealHumphrey:
@ostermanj and @ptgott - is it safe to fully delete the `/javascript` folder since the central version is now moved over to docs?

ptgott:
@ostermanj @NealHumphrey 
Thanks for wrapping up, @ostermanj!
I can update PR #269 to complete the menu gray-out code after this weekend if it's useful to do so (depending on what happens with integrating the filter view). Re: the order of selecting menus, the idea with the code would be that selecting an option in one of the menus would gray out unavailable items corresponding to that option in the other one (kind of like an online clothes shop where clicking one color grays out unavailable sizes and vice versa). As you can see, it doesn't work right now, but that was the idea.

I'm pretty sure '/javascript' is redundant now! If we need it, we can copy it from an older commit.

docs/tool/index.html
```diff
@@ -1,334 +1,60 @@
...
-		<script src="https://code.highcharts.com/modules/data.js"></script>
-	<script src="https://code.highcharts.com/maps/modules/data.js"></script>
-	<script src="https://code.highcharts.com/maps/modules/exporting.js"></script>
-	<script src="{{ site.baseurl }}/js/external/papaparse.min.js"></script>
-
-	<script src="{{ site.baseurl }}/js/maps.js"></script>
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Sandbox option 2</title>
```
NealHumphrey: Should be Housing Insights Tool or something

docs/tool/index.html
```diff
@@ -1,334 +1,60 @@
...
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/d3/4.8.0/d3.min.js"></script>
+    <script src='https://api.mapbox.com/mapbox-gl-js/v0.32.1/mapbox-gl.js'></script>
+    <script src="js/d3-tip.js"></script>
+    <script src="js/PubSub.js"></script>
+    <script src="js/chart-prototype.js"></script>
+    <script src="js/map-view.js"></script>
```
NealHumphrey: I don't see rich-map.js in this page. Am I correct that rich-map.js is deprecated and can now be deleted from `/docs/tool/js`?

docs/tool/index.html
ostermanj: Yes -- should be ok to delete old map-view.html and the javascript/ folder. I'll update the title.

ostermanj:
That's right -- rich-map is deprecated.

PR #275, opened 2017-05-20
Song dev2

terrysky18:
Finally finished development for this bug fix. Here is the list of changes:

- Added 'checkTable' function to check whether the new table is in table_info.json
- Added 'appendJSON' function to append the new table to table_info.json
- Added 'duplicateTable' function to handle when the new table is a duplicate entry in table_info.json. User has the option either overwrite the entry or cancel.
- Added a new command line argument 'single-add' to handle the new process

TODO:
- Create backup of 'table_info.json' (meta.json) before updating the file; Version Control
- Unit test script of the new functions


terrysky18:
This pull request is a fix to Issue#196 Add functionality to make_draft_json for adding new sources.

PR #279, opened 2017-05-21
Filter module with first slider

NealHumphrey:
So I didn't get to do any coding at the hackathon (helping everyone else!), and it also seemed like it would be useful for me to go ahead and set up a home in the code for @laurkgol's categorical picker and @MaryColley's date pickers. 

This is a pretty good start at setting up a demo of our 3 filters, and done in a way that it can be easily expanded to an arbitrary list of filter criteria. I'm currently publishing the state change of my one slider just as a single 'filterValues' so we still need to handle how to wrap all the filter criteria up into one variable. 

- Adds a temporary sidebar for filtering
- Adds a slider that can publish its changes to the state module
- Sets up structure for inserting all filter ui elements based on config data

Other stuff:
- Shrinks sidebar-view temporarily to make development easier (this sidebar will either get re-arranged or put into a slide in-out thing)
- adds import of Slide.js which is one option for making data navigation components navigable. But, doesn't implement this yet (left it in to help remember what looked like best pick after research)

Here's a demo:
demo_slider


NealHumphrey:
@ptgott or @ostermanj it would be great if you could give this PR a test and review!

@MaryColley and @laurkgol there are spots in the filter-view.js file reserved for you to put your ui components in ("//set up categorical pickers" and "//set up date pickers") once this gets merged to dev

ptgott:
Looks like a good start! I can review Monday evening unless @ostermanj wants to do it sooner.

ostermanj:
I agree it looks great! Very cool that the framework is there to know programmatically which type of filter to add in for a specific data set.

The PubSub module allows adding subtypes to messages dot-notation, so we could handle publishing changes to the state of multiple filterable fields like this:

`filterValues.<field>, [<min>,<max>]`

So the project count would be `setState('filterValues.proj_units_tot', [5, 200]`, for example. A function subscribed to "filterValues" would be called whenever any filterValues subtype stateChange was made, but you can also have functions subscribed only to the specific subtype by using that dot-notation in the `setSubs` method.

I guess the subscribed functions may have to know what kind of filter (continuous, date, categorical) was triggered. They could look that up based on the field, or it could be passed in as in `setState('filterValues.proj_units_tot, {continuous: [5, 200]}` or some other way, if necessary.

Small change to previous commits: in line 91 of map-view.js, `if ( existingOverlayType !== overlay.proj_name )` should be `if ( existingOverlayType !== overlay.name )`. 

I think worth approving with that small change (may have been my typo on earlier commit), but let's wait for @ptgott 's review too.

NealHumphrey:
@ostermanj Oh, that's great - I didn't realize that the pubsub dot notation worked like that, I thought you'd just used the dot-noted keys as a convenient way to group similar names. That will make creating the real filter state setup (which I left as a TODO) much simpler. 

The first part of my code has the `filterView.components` variable. I think that is going to be coming from the API instead of hard-coded. That is the DRY source of truth of what type of variable a particular field is (categorical, continuous). As such, the code that monitors the filterState can probably just look it up directly from there instead of embedding it into the setState.

NealHumphrey:
@ptgott  and @ostermanj I merged this with a) the line that john noted fixed, and b) changed to make the state module accurately record which slider moved. But, in b) I had to use the ES6 'let' variable. We should discuss this tonight, and whether we should stay w/ es5 which would require a constructor.

PR #280, opened 2017-05-21
Added filter module at utils/filter.js

wassona:
I started a filter module that responds to a 'filterChange' publication.

In housing-insights.js, I replaced some slightly awkward .className calls with .classList calls.

This also adds Laura's multi-selector sandbox element


NealHumphrey:
@ostermanj I'm merging this in even though it's a work in progress - it doesn't interfere with current code, and I'd rather have all our code in one place for tonight.

NealHumphrey:
@ostermanj if you want to do a code review on it and suggest any edits, we can either discuss in person w/ @wassona or you can put comments on it here even after the PR is closed.

PR #282, opened 2017-05-21
257 refactor load data

jkwening:
Refactor - encapsulated load_data.py into a class but still honored command line feature. As part of the refactor I added code to deal with the odd case that a user provides multiple databases as a choice for loading - it defaults to loading to one with minimal impact.

Note - there was a merge conflict related to new code added to the 'drop_table' function in dev compared to what I had in the 'drop_table' method. I was able to resolve the conflict by moving the additional code into the 'drop_table' method per encapsulation.


jkwening:
Closing this PR because this was included in PR 300 when it was merged into dev.

Issue #274, opened 2017-05-20
Opendata.dc.gov connection class


NealHumphrey:
Opendata.dc.gov provides multiple data sets that we need. They provide an api endpoint for each data file that we need already in .csv format. 

In our `/python/housinginsights/sources` folder there is a base.py file with a BaseApiConn class. 
- [x] Create a new class called OpenDataApiConn() (in a new file called opendata.py) that inherits from the base class. 
- [x] add properties to __init__ with an appropriate path to where to save each data type (e.g. /data/raw/tax or something like that)
- [x] add a get_XXX method for each of the data sources (where XXX is the name of the data source)
- [x] use get_XXX to download the csv file and save it to the appropriate sub folder of `/data/raw` (e.g. `/data/raw/tax` or `data/raw/other`
- [x] add a get_data method that calls all of the get_XXX methods.  

To start, get this data download class to work with two data sets:
- building permits 2016 http://opendata.dc.gov/datasets/building-permits-in-2016
- tax public extracts tickets http://opendata.dc.gov/datasets/integrated-tax-system-public-extract-facts

NealHumphrey:
Look at the /python/housinginsights/sources/mar.py for an example of inheriting from the base class, though this does not use the csv method

ajhalani:
Relates to https://github.com/codefordc/housing-insights/issues/187 as well.

ajhalani:
Please note I wrote the base opendata api conn class in https://github.com/codefordc/housing-insights/pull/283. Would love feedback to avoid duplicate work.

Creence:
I already wrote this yesterday and did not have a chance to check it in.

Creence:
@ajhalani I added mine because the get_data method is supposed to call all of the get_XXX methods in #284 and yours seems to be a generic function for getting a particular file.  This ticket was assigned to me yesterday when I worked on it.  If you want to work on something you need to assign it to yourself to avoid duplicate work.

ajhalani:
Ofcourse. Not my intention to do someone story assigned to someone else :). SInce story is assigned to you, I defer to your PR. My crime data story needed the base class so thank you! 

@NealHumphrey  - The way the requirements are written for this story, wouldn't opendata logic for pre-defined filepaths be inconsistent with existing get_api_data/mar.py. mar.py which take path of file as a paramter --output.
Also 
> add a get_data method that calls all of the get_XXX methods.
[...]

PR #284, opened 2017-05-22
python/housinginsights/sources/opendata.py

Creence:
Opendata.dc.gov connection class #274 according to directions in issue.

### DIFF
--- a/python/housinginsights/sources/opendata.py
+++ b/python/housinginsights/sources/opendata.py
@@ -22,7 +22,7 @@ class OpenDataApiConn(BaseApiConn):
 
     # This function is used as a decorator
     def opendatas(self,get_data_func):
-        get_datas.append(get_opendata_func)
+        get_opendatas.append(get_opendata_func)
         return get_opendata_func
     
     @opendatas
--- a/python/housinginsights/sources/opendata.py
+++ b/python/housinginsights/sources/opendata.py
@@ -1,0 +1,61 @@
+from requests
+
+from housinginsights.sources.base import BaseApiConn
+
+class OpenDataApiConn(BaseApiConn):
+    """
+    Opendata.dc.gov provides multiple data sets that we need. They provide an api endpoint for each data file that we need already in .csv format\
+    To add new datasource, go to the opendata.dc.gov site.
+    In the API tab, change the URL in geojson to end in .csv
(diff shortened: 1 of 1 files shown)


Creence:
Yes, that is what I meant.  Good catch.  I was rushing to do the pull
request.

ajhalani:
Would it be possible to add example of how this would work with get_data_api.py ?

python/housinginsights/sources/opendata.py
```diff
@@ -0,0 +1,61 @@
...
+    def get_tax_public_extracts_tickets(self,save_name='integrated-tax-system-public-extract-facts.csv'):
+        """
+        Download csv folder and save to appropriate subfolder
+        tax public extracts tickets 
+        http://opendata.dc.gov/datasets/integrated-tax-system-public-extract-facts
+        """
+        url = 'https://opendata.arcgis.com/datasets/014f4b4f94ea461498bfeba877d92319_56.csv'
+        resp = requests.get(url)
+        url_save_name = os.path.join(self.data_type_save_place, save_name)
+        f = open(url_save_name, 'wb')
```
ajhalani: Base class has a function for writing to csv file. Any value in putting this another function to write to file in base class, so potentially other ApiConn could leverage too ?

python/housinginsights/sources/opendata.py
```diff
@@ -0,0 +1,61 @@
...
+        f.write(resp.content)
+        f.close()
+
+    @opendatas
+    def get_building_permits_2016(self,save_name='building-permits-in-2016.csv'):
```
ajhalani: Wondering how this would extend to previous years (backfill), as well as future 2017 and ahead..Would it make sense to have a year parameter?

NealHumphrey: @ajhalani yes, I agree it seems to make sense to have a year parameter. But the problem is the way opendata.dc.gov handles this is each year is a new 'dataset' and we ESRI assigns a random string to their url endpoint, so there's no way to know what the 2018 endpoint will actually be until they make it. No matter what we're going to need to edit the code to allow us to get next year's data.

NealHumphrey:
I merged this code into a master 'combined-api-prs' branch with the other related branches and made a new pull request for the merged branch.

PR #283, opened 2017-05-21
187-Created common opendata conn class and implemented crime download

ajhalani:
This should also set the base for https://github.com/codefordc/housing-insights/issues/274, impacts https://github.com/codefordc/housing-insights/issues/198, https://github.com/codefordc/housing-insights/issues/194,  and other issues for data from opendata. 

Sample usage - 
`python get_api_data.py opendata get_crime --output ../../data/raw/crime/2017.csv --params "year:2017"
`
screenshot_2017-05-21_15-21-56

Please note csv files from opendata come with BOM, so we might still need to test with any cleanup logic that was written.

PR #278, opened 2017-05-20
#198 download tax data

melanatech:
I don't know why it's merging 69 commits, but there should be some additions to tax.py and base.py now.

Issue #260, opened 2017-05-12
API save location as a property and create folder if needed


NealHumphrey:
The API `XXXApiConn()` classes are inconsistent in how they choose where to store the output csv file created from their data source. `TaxApiConn` (about to be merged in) added the save folder as a hard-coded relative path, whereas others accept a user input. All of them also fail if the folder does not already exist. 

In preparation for automatic data downloading, we should unify this with a consistent method. 

- [ ] add a `_save_folders` property and `_file_names` to the `BaseApiConn` class. Make them null, so the child class will override. This will be dictionaries so that we can handle when there are multiple output files in a class. 
- [ ] add an `output_path` (using the `@property` decorator) to the `BaseApiConn` class that stores and returns the output file by joining the above properties. This should have custom code (hence the @property decorator) that resolves relative file paths, adds a datestamped folder, and creates the folder if it does not yet exist. 
    - probably a good idea to pass a tuple of  (_save_path, _file_name) to the setter. _save_path should be to the /data folder, for example `raw/tax/opendata` instead of `../../data/raw/tax/opendata`. _file_name can be something like `tax.csv`. [Here's an example](http://stackoverflow.com/questions/18714262/property-setter-with-multiple-values) of passing an iterable to a setter. 
    - Use the `__file__` attribute to get the relative path of the base.py file to resolve the relative path to /data, so that it works no matter what folder the code is run from. 
    - If _save_path or _file_name are null, return null

- [ ] modify the the result_to_csv to check for the `output_path` property; if it is null, raise an error. This way new developers will know they need to add this property to their apiconn. Also allow result_to_csv to receive an optional override path to save the file. 

Goal is to have the `output_path` property return something like: `C:/path/to/housing-insights/data/raw/tax/opendata/20170502_163215/tax.csv` where 163215 is hhmmss format (so we avoid overwriting data if we run it twice in a day)

NealHumphrey:
@jkwening I'm thinking about this approach for our API save paths, as a first step towards automatic api data pulls. I ended up being pretty detailed in the ticket - I can either implement or if you want to do it after you finish the phase 2 of the DCHousing feel free. Whoever starts on it first post up here to claim.

NealHumphrey:
Also available for anyone else that wants some ingestion-related work and wants to learn about @property techniques!

xee5ch:
[...]

Issue #187, opened 2017-04-10
Crime data via the opendata API


NealHumphrey:
We currently have crime data that is manually downloaded from the Crime Data Map. 
Opendata.dc.gov provides the same data via an API, and with extra geocoding data added to it (lat/lon etc.). It looks like the newly-updated opendata.dc.gov is now being updated daily with this info, so we should switch sources so we can keep getting new updates. 

- Create a CrimeApiConn class based on the MarApiConn class currently in the code. Use this to add raw data files to `/data/raw/crime/opendata/YYYYMMDD` when  the `python/cmd/data.py` script is run with the appropriate arguments. Write a demo command to run at the command line to create this file with the appropriate datestamp. 

- Run this for all of the past years of data (each year is a separate "dataset" on opendata.dc.gov). Follow the '[add a new data source](http://housinginsights.org/resources/adding-data-sources.html)' instructions. 

Future: automatically add new files to the pipeline whenever this update script is run, but we wait to do this until we have created an appropriate structure for all API scripts.

just4jin:
in progress

jasonrhaas:
I'm interested in improving the data ingestion automation process as well.  This issue and #198 seem to be related.  Both are using opendata.dc.gov, so I think that rather than generating custom data ingestions for each, we should have a more generalized approach for ingesting data from opendata.dc.gov.

I think that using the API is 100% the right approach, rather than manually downloading data.  Even better would be using a generalized python module for pulling data, to abstract out the differences between data sources.

For example, in the code we might have:

```
from dcdata import Crimes, Taxes

crimes = Crimes.pull()
taxes = Taxes.pull(year=2012)
```

And of course we can manipulate the data from there to get it into our future SQLAlchemy format.

NealHumphrey:
@jasonrhaas Yes, it would be good to use the same approach - though part of breaking things up was to have things that multiple people could work on it once. 

@jkwening is leading up the Python side of the data ingestion code so he's the best one to coordinate with!

NealHumphrey:
@ajhalani and @just4jin flagging an issue for you that came up in the tax ingestion. The json api's have a default limit of 1,000 records. Looks like the best solution is using the 'geojson' api endpoint, but swap .geojson for .csv to get a bulk download of all the api already in csv format.

ajhalani:
Thanks Neal.

ajhalani:
Work completed in PR - https://github.com/codefordc/housing-insights/pull/286

[...]

PR #286, opened 2017-05-23
Combined ApiConn pull requests

NealHumphrey:
This pull request combines code from #278, #283, and #284, and then adds on to it to standardize the method signature of the ApiConn classes. It also replaces the get_api_data.py with a new approach. This resolves issues #274, #187, #260, and #198. Issue #16 is included but still outstanding because we are waiting to find out when opendata.dc.gov will update their 2017 building permits data set. 

Overview of changes in this PR:
- Base class has _available_unique_data_ids list which needs to be included in all classes. Even if there is only one id it should be a list, to ensure consistent behavior between classes that produce one file vs. those that produce multiple. 
- Base class now has an `@property` called `output_paths` that returns a dictionary of `unique_data_id:absolute_local_output_filepath` for all the _available_unique_data_ids. This automatically builds in datestamping. 
- All files are saved to the same datestamped folder using their unique_data_id + ".csv" as the filename. This should simplify the code that needs to manage how to update the manifest.csv to reflect updated data sources - every datestamped folder can be searched for the unique_data_ids that it has in it, and the load_data code can only update the data for the files that it finds. 
- Every ApiConn class now should have a `get_data()` method. It should be able to run without any mandatory parameters, and it must contain the following (user) optional params: unique_data_ids (to allow for flexibility of not always downloading all data sets), `sample` (for easier testing by downloading only a few rows), and `output_type` (normally should be 'csv' but can be 'stdout'). 
- Cleaned up the opendata.py class so that it is not necessary to inherit from base class, new 'normal' data sources (those that don't need special processing) can be added just by adding them to the list. 

@jkwening Can you review this?
@ajhalani Check this out, maybe you can do a code review as well? It doesn't look like it because I had to rewrite a bunch of stuff so to work with the new approach to get_api_data.py but your code was a very useful influence.


NealHumphrey:
@emkap01 Heads up that when this PR is approved it means that the crime, building permits, and tax data sets are all done with the 'write api code' step and need to move on to the 'cleaner/ingestion' part of the process.

ajhalani:
👍

PR #287, opened 2017-05-23
Add initial checkin of TOPA notices for 2015-2017.

xee5ch:


xee5ch:
Sorry for the dealy.  This is to address #287.  I had to rebase and clean up after massacring files with IDE auto-cleanup of whitespace.  Did not want to pollute the tree and waste your time.

PR #290, opened 2017-05-24
Small additions to the filter logic

NealHumphrey:
@ostermanj I was thinking about the ES6 issue this morning and took a crack at it. Turned out to be pretty easy to solve with a closure - I think the first time I've every actually understood how a closure works and needed to write one. So that's cool. 

While I was at it I went ahead and fixed up the state change events so that the util/filter.js code would trigger based on filter events. Looking at the logic with it hooked up though I realized we'll need to re-write most of the code so that we can use the same system of building up the logic based on the filter controls that have been added to the list, instead of hard-coding variable names. So for now I commented out the filter logic. @wassona we should discuss if it's best for you or John or I to take a stab at rebuilding the filter logic. 

@ostermanj can you do a quick review of this and merge if you're ok w/ it?

Change list:
- Fixed the ES6 issue w/ publishing state changes to filtering by using a closure
- Hooked up the /util/filter.js file to the filterValue state change
- Commented out the current incomplete filter method
- Added a demonstration list of filter values to show that deduplication util method works correctly
- Added 'data_type' to the components config data so that the util/filter.js will be able to distinguish between decimal and integer fields (since both get a continuous slider but filter logic needs to be implemented slightly differently)

### DIFF
--- a/docs/tool/js/util/filter.js
+++ b/docs/tool/js/util/filter.js
@@ -1,10 +1,13 @@
-var filter = {
+var filterUtil = {
+	
 	init: function(){
 		setSubs([
-			['filterChange', filter.publishFilteredData],
+			['filterValues', filterUtil.publishFilteredData],
 			// need to subscribe to data load and run a save the data
 		]);
 	},
+
+
 	options: { // TODO replace with a function that populates this according
 				// to data from the API
 		ward: '',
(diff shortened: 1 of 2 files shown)


ostermanj:
Ah, closure . . . see if you can remember how it works after a couple days of not thinking about it :)

I'm cool with merging. There is still a bit to figure out about how to handle the filtering, but it may be simpler than it seems. Like do we need an `options` object in filter.js? The filter UI elements are based on the data and should be able to communicate everything needed for a filterValues state change to trigger the functions to update the map display or a d3 display. Seems likes util/filter.js
is trying to reproduce some of the logic already baked into filter-view.js, if that makes sense.

Of course I could be missing something . . .

NealHumphrey:
@ostermanj I agree that we probably want to remove the `options` object as well - this was created before filter-view.js was created, and when I made filter-view I tried to do as reusable approach as I could which we hadn't talked through as a design feature of the filterUtil.

Issue #272, opened 2017-05-19
Uncaught exception when going from building view back to map view. 


NealHumphrey:
@ostermanj to investigate later, doesn't appear to break any functionality. This applies to the refactored code versio in the current pull request #270 soon to be merged to dev. 

To recreate:

Go to /tool page
Navigate to a building view (I went to /tool/building.html?building=NL000360)
Click back button. 

This appears in console:
```
PubSub.js:55 Uncaught TypeError: Cannot read property 'length' of undefined
    at t (https://cdnjs.cloudflare.com/ajax/libs/d3/4.8.0/d3.min.js:5:23198)
    at Object.entries (https://cdnjs.cloudflare.com/ajax/libs/d3/4.8.0/d3.min.js:5:23683)
    at DonutChart.returnPieVariable (http://127.0.0.1:4000/tool/js/donut-chart.js:18:8)
    at http://127.0.0.1:4000/tool/js/sidebar-view.js:87:43
    at Array.forEach (native)
    at changeZoneType (http://127.0.0.1:4000/tool/js/sidebar-view.js:86:28)
    at callSubscriberWithDelayedExceptions (http://127.0.0.1:4000/tool/js/PubSub.js:61:13)
    at deliverMessage (http://127.0.0.1:4000/tool/js/PubSub.js:82:17)
    at deliverNamespaced (http://127.0.0.1:4000/tool/js/PubSub.js:93:13)
```

Nothing appears broken on page so might be non issue.

NealHumphrey:
I tried doing a bit of digging to fix this. Couldn't quite trace it, but it appears to be related to sequencing of when the data is available vs. when this function is called.  the `chart.data` object is empty when changeZoneType is called. 

sidebar-view.js sets these three subscriptions:
```
setSubs([
            ['firstPieReady', sideBar.setDropdownOptions],
            ['mapLayer',sideBar.changeZoneType],
            ['pieZone', sideBar.changeZoneType]
        ]);
```

My guess is that if mapLayer finishes before firstPieReady it triggers the changeZoneType event but the data is null. 

This probably isn't a big deal since the user won't be able to alter the ward until the firstPieReady event happens. So maybe we want to just handle this with some sort of error handling where if chart.data doesn't exist it doesn't raise an error unless the firstPieReady event has occured? 

[...]

PR #291, opened 2017-05-24
mapView and buildingView as single page app

ostermanj:
This changes how you switch from one view to another, treating it as a single page app where the existing view fades and is stashed off screen while the new view slides in. The new view is appended from a partial html file. It's working now for the map view and building view.

On initial load, index.html  basically a shell without content. Map view partial is appended in and its `init()` function is called as part d3.html's success function. The same happens when the building view is made active for the first time. `init()` functions are called only the first time a view is made active; anything that needs to happen every time should be in the view's `onReturn()` function.

The data persists through the view change, so the building view would make use of the existing state; going back to the map view finds it's state (filters, overlays, etc) still in place.

TBD: do we need URLs to reflect changes in views or the state for version 1. If so, that'll take some doing.

Relevant to issues #272 and #212.

@NealHumphrey @ptgott @wassona : grateful for review / thoughts


ptgott:
@ostermanj I like the sound of a single-page architecture. We’ve definitely talked about this as a way of persisting data across view changes, so it’s cool that you’ve gone ahead and made it happen. I haven’t reviewed it yet, but can get to it by tomorrow afternoon.

Issue #254, opened 2017-05-10
Investigate DCHD TOPA Monthly Filings for Ingestion


xee5ch:
Per conversation with @jkwening and the rest of the Browse/Ingestion Team, there is a high demand for TOPA data.  According to the DC website, they have monthly updates going back to 2016 listed below.

https://dhcd.dc.gov/page/weekly-report-tenant-opportunity-purchase-act-topa-filings

Each month links to a subsequent page with attached PDF, which appears to be a PDF export for from Excel directly.  Not sure the reliability, feasibility, and viable options for data massage and ingestion.  Will investigate further as part of #237 push.

xee5ch:
Per @emkap01 there is a sample XLS file that seems to be a cleaned version of similar data that can be found.  Just leaving this for future reference or analysis.

https://docs.google.com/spreadsheets/d/1zRGbvcg2X6RrwLuDdTq9SmCXQo-Mp_GOO-9kuxYc60Y/edit#gid=0

xee5ch:
Minor progress and early in the process.

- Tested building a Docker container with normal CPython 3 and pdfquery (because I cannot get it working with Anaconda since it is not the default channel and is a per user build)
- Emailed DC OpenData address like Neal in other tickets asking if the only way to get access to this is through PDF.

@NealHumphrey: TOPA data is marked as high and PDF of Excel looks like the only source, how bad do we want routine manual massaging of the data to get it in?

NealHumphrey:
@xee5ch CNHED maintains (maintained?) a Google Spreadsheet where they manually pulled out some of this information - which could be a source for historical data, at least. I'm not sure how up to date it is. Besides that, I think that getting this in machine-readable format (probably an XLS spreadsheet) is possible but will require pressure to DHCD. The best method to do that is
a) Contact opendata, the open data office (Traci Hughes), and DHCD to start the process of pushing for it in machine readable format instead of just pdf
b) Go ahead and set up some PDF reading to require manual intervention, plan to update our data maybe once a month or so, show our tool to DHCD and use that to say hey, it actually is important. 

If we don't get DHCD cooperation on it, we can decide whether to keep doing any manual processing of PDFs, but for now it's better to get it in even if we're not sure we can maintain it. 

I'm copying you on an email to Scott at CNHED to see if he knows more.

xee5ch:
[...]

Issue #194, opened 2017-04-12
DC-funded affordable housing units on opendata.dc.gov


NealHumphrey:
Add the DC-funded buildings to our list of all affordable buildings and subsidies using the [Affordable Housing dataset](http://opendata.dc.gov/datasets/34ae3d3c9752434a8c03aca5deb550eb_62) on opendata.dc.gov. See notes on what is in this data set in #137 . 

**Current related data:**
Our current database contains a 'projects' table, which is pulled from the [Preservation Catalog](www.neighborhoodinfodc.org/dcpreservationcatalog/). In addition, it has a 'subsidy' table, where project -> subsidy is one->many.  This catalog aggregates all the subsidized buildings from the lists published by the federal government; therefore it only contains buildings federally subsidized buildings. It does not systematically include buildings that have only received subsidies from the DC Government; 

**Data adding**
We want to create a row in the project table and in the subsidy table corresponding to each additional building in the opendata.dc.gov dataset. 

There are several ways that a duplicate record could be created, which we want to avoid:

- LIHTC funding is distributed by DHCD; the PresCat includes data from the federal list of LIHTC recipients, so we would expect most of these buildings to be duplicated (and would like to flag any that are not in the federal list but are in the DHCD list. 
- A federally subsidized building can get an additional subsidy/rehab loan from DHCD. In this case, we want to add a record only to the 'subsidy' table and link it to the proper 'project' row using NLHIC_ID
- Occasionally buildings are manually added to the PresCat, so these may already be in the catalog. We want to update any fields in the 'project' table if better data is available from DHCD. 
- There a large number of subsidy records (and corresponding project records) that have the subsidy source "DC/AFFPIPELINE". These are at least 8 years old (though the data for the subsidy is likely still valid) - there may be a few overlapping buildings but for the most part new buildings are not being added to the Preservation Catalog currently. 

Methodology
- [x] Create a DCHousingApiConn class based on the MarApiConn class currently in the code. Use this to add raw data files to /data/raw/tax_assessment/opendata/YYYYMMDD when the python/cmd/data.py script is run with the appropriate arguments. Write a demo command to run at the command line to create this file with the appropriate datestamp.
(pull request after step 1)
[...]

Issue #136, opened 2017-03-20
Scrape the DHCD DFD Pipeline public database and add it to our ingestion pipeline


NealHumphrey:
DHCD provides a public dashboard of the projects they are funding along with their current phase of development. This is a quickbase application - [browsable version here](https://octo.quickbase.com/db/bit4kvfmq). From our interview with Chris Dickerson-Prokopp it sounds like this data is probably pulling from the most up to date data (we will want to verify this eventually). 

Quickbase provides API methods to access the data. It returns XML. Unfortunately the database format does not appear to be documented by DHCD (based on a quick look - also something to verify). We can probably get information from DHCD by reaching out; or, we can guess the appropriate API calls by using the online browsable version. 

This API call, for example, returns one building record with its loan amount and status:
`https://octo.quickbase.com/db/bit4krbdh?a=API_DoQuery&query={3.EX.1}`

The generic quickbase API_DoQuery is [documented here](https://help.quickbase.com/api-guide/do_query.html#queryOperators). 

- [ ] Play with the API field query parameters to figure out what we can obtain from the database and make a list. 
- [ ] Write a Python script that can extract records for all properties in the database. We especially want to know loan amount, project name/address, and status - but we should save all data available. This should save data into a folder here: `housing-insights/data/raw/dhcd_dfd_pipeline_db/YYYYMMDD` with the datestamp of the date it was scraped. 
- [ ] Integrate this with the building-matching script that is a WIP by @NoMoreOreo (issue #107 ). Use this to add the nlihc_id to any buildings we already have from the PreservationCatalog. 
- [ ] Add the most recent version of the scraped file to the manifest.csv, the meta.json and add the appropriate cleaner so that this can be ingested into the database. 
- [ ] Meanwhile, reach out to DHCD and find out a) if the data is live-updated, b) obtain the data dictionary that we can use to construct the appropriate API calls without guessing, c) find out what data is not included in the public database and if we can get access to it. Neal is starting this via Chris Dickerson-Prokopp who sent us the database link; I'll add whoever takes this ticket to the email discussion.

NealHumphrey:
[...]

PR #293, opened 2017-05-25
194 dchousing injection

jkwening:
@NealHumphrey This is a rudimentary solution to #194 and will require refactoring so the code is reusable for #136 and #254 with regards to populating nlihc_id field but wanted to get the first pass out to everyone else. The pertinent functions are inbedded in insertion_functions.py along with test cases.

The current workflow is to:
1. clean up project.csv to make sure all rows have a valid 'ADDRESS_ID' which is used as the lookup and mapping to nlihc_id.
2. create a mar_proj.csv that maps all existing nlihc_id to the mar data for the address id
3. create dchousing_append_project.csv and dchousing_append_subsidy.csv by doing a lookup against mar_project.csv and creating uuid values for nlihc_id if a building in dchousing isn't already in the project table.

**On Tuesday we discussed adding 'MARID' field to the project table since I found some incorrect values in 'Proj_address_id' and also trying to get the complete mar data in single batch. I think I've found a way to do the latter via opendatadc.gov page. I still think we need a mapping of nlihc_id/uuid included in the mar file so that we don't continuously generate new uuids each time we reload the data - something we can discuss further**

Notes: modified subsidy and project related cleaners  to handle null values

TODO: need clean dchousing dtm before loading into database - I already have custom code for doing that but it should be called before creating the project/subsidy append csv files.

Sorry for the long write up - wanted to make sure I got some of the next actions steps we discussed written down somewhere; including any outstanding issues that were still on my mind.


jkwening:
Oh forgot to add - I refactored make_draft_json.py to take command lines so we don't have to manually go into the file and fill in the needs arguments.

PR #295, opened 2017-05-27
Minor changed to wmata_distcalc.py

jgordo04:
Added some formatting for the stdout output format


jgordo04:
Replaced some print statements with logging.info

python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -133,10 +134,11 @@ def get_dist(self, unique_data_ids=None, sample=False, output_type='csv',db='loc
                 self._array_to_csv(self.distHeader, self.distOutput, self.output_paths[u])
```
ajhalani: Unrelated to changes in this PR but caught my eye, maybe in future we can reuse result_to_csv in parent base class. https://github.com/codefordc/housing-insights/blob/dev/python/housinginsights/sources/base.py#L128

ajhalani:
Don't have merge access but approved the changes.

jgordo04:
Added option to output wmata_distcalc.py results to an alternative file. Added unit test for wmata_distcalc.py

python/housinginsights/sources/wmata_distcalc.py
NealHumphrey: Ya, the _array_to_csv expects a different format (list of lists, where each inner list is a row i.e. `[[a,b,c],[x,y,z]`)  from result_to_csv (json, i.e. '[{'col1':'a','col2':'b','col3':'c'},{'col1':'x','col2':'y','col3':'z'}]`). We should probably combine all these into one method (or several) and put them all on the base class.

PR #298, opened 2017-05-29
212: Integrate the building page into the front-end refactor

ptgott:
For issue #212

**Squashed commits:** Move hardcoded data URLs from old building view code into the new buildingView object in building-view.js. Move remaining building-map.js and building-header.js code into building-view.js. Move obsolete docs into new 'delete-after-test' folder for until we're sure we won't need any code from them. Get subsidy-timeline.js to match the recently refactored Chart constructor. Persist selected building data from the map view to the building view through setState.


NealHumphrey:
@ptgott I see you pushed another commit, let me know when this is ready for a review! I'll assume it's still a WIP until I hear otherwise.

ptgott:
@NealHumphrey @ostermanj 
Here's my wrapped-up-for now work on integrating the building page with the refactor. As you'll see, there's close to no styling, so we'll need to do some work on the look and feel. The goal has been to get the pre-refactor work to display again, not to revisit any fundamentals, so we'll also have to work out the precise information to show and how to show it.

Another thing: I've removed all inheritance between ChartProto and SubsidyTimelineChart. This is because the 'initialize' method for the former had some functionality that wasn't really appropriate for the latter. Rather than change the parent constructor, I opted to have SubsidyTimelineChart use only its own 'initialize' method. I know we've talked about implementing the recommendations in [this](https://bost.ocks.org/mike/chart/) article, so if you'd like we can set up a new issue for reworking chart inheritance.

ostermanj:
That all sounds sensible. I can probably review tomorrow unless you get to it before me, @NealHumphrey

NealHumphrey:
@ptgott yes sounds good to me! 

On the article you mention, we're using a slightly different approach to chart inheritence since we went down the prototype path instead of the closure path, which I think are a little bit at odds with each other? In any case, we can leave chart inheritance code structure for its own discussion outside of this, and focus on using the method @ostermanj set up. 

@ostermanj it would be great if you're able to review today! Ping me if you don't think you'll be able to get to it before ~4 or so, though, and I'll take a shot.

ostermanj:
I should be able to start looking by 4.

ostermanj:
It's working great, with styling and data decisions to come, as Paul describes. It has the functionality from before the refractor, making use of the state changes to persist the selected builfing from the mapView to the building view. All good and will approve.

My latest PR #310 will have small conflicts with this after it's merged, but they're minor. Basically defer to Paul's conflicting changes in building-view.js and mine in map-view.js, with the exception that Paul's `setState('selectedBuilding', building);` will need to part on the popUp's click event.

PR #300, opened 2017-05-30
132 combo replace append

jkwening:
Added new method: update_only() to the refactored class encapsulation of load_data.py. This method allows for updating only the data for a specified set of unique_data_ids. Included in this pull request are test cases that verifies that no duplicates are generated.

Note: 'replace_table' field in meta.json now seems obsolete because for any given table we can specifically update each unique_data_id that composes the table.

Outstanding:
- clean up manifest to make sure unique_data_id stays the same across different versions of the same data set (i.e. if the tax data is updated, id is still 'tax')
- make sure there is only every one unique_data_id per file that is in the database at one time

Thought:
Since we're now consistently using one unique_data_id for each data source - we can add a method to the get_api_data that automatically updates the manifest.csv whenever a new batch is pulled. Concern - how do we effectively validate that the new raw data is ready to be processed?


python/scripts/load_data.py
```diff
@@ -20,208 +20,403 @@
...
+                csv_writer.remove_file()
+
+    def _update_database(self, table_name, sql_interface):
+        """
+        Load the clean PSV file into the database
+        """
+        print("  Loading...")
+
+        # append new unique_data_id data instead of dropping entire table
+        # TODO: remove unnecessary code - relace table not needed?
```
NealHumphrey: Yeah, we can remove the replace_table stuff. Instead, we'll tell it whether or not to replace the unique_data_id (i.e. by deleteing the rows).

python/scripts/load_data.py
```diff
@@ -20,208 +20,403 @@
...
+                temp_filepath = self._get_temp_filepath(
+                    manifest_row=manifest_row)
+
+                # get objects for interfacing with the database
+                sql_interface = self._configure_db_interface(
+                    manifest_row=manifest_row, temp_filepath=temp_filepath)
+                sql_manifest_row = sql_interface.get_sql_manifest_row()
+
+                # if data_id has status = loaded in sql_manifest
+                try:
```
NealHumphrey: I think that to combine load_all_data with update_only, we'll want to put all the stuff in this 'try' block into a `_remove_existing_data` method and wrap that in an if statement to only run if the user says to update the data. 

In addition - the best way to handle this is probably file by file. If we add a column to the manifest ('delete_existing') then that can be the flag as to whether to delete or not.

python/scripts/manifest.csv
```diff
@@ -70,3 +69,6 @@ superseded,wmata_dist,wmata_dist_20170215,2/15/2017,latin-1,../../../data,https:
 superseded,wmata_info,wmata_info_20170215,2/15/2017,latin-1,../../../data,https://s3.amazonaws.com/housinginsights/,raw/wmata/20170215/wmatainfo.csv,
 use,wmata_dist,wmata_dist_20170215,5/24/2017,latin-1,../../../data,https://s3.amazonaws.com/housinginsights/,raw/wmata/20170215/dist_updated_headers.csv,
 use,wmata_info,wmata_info_20170215,5/24/2017,latin-1,../../../data,https://s3.amazonaws.com/housinginsights/,raw/apis/20170524/wmata_stops.csv,
+____DATA FOR APPENDING____,,,,,,,,
+use,project,dchousing_project,5/5/2017,latin-1,../../../data,https://s3.amazonaws.com/housinginsights/,raw/preservation_catalog/20170315/dchousing_append_project.csv,append to 20170315/project.csv
```
NealHumphrey: @jkwening Can you upload these files to S3? I couldn't run them.

python/housinginsights/ingestion/Cleaners.py
```diff
@@ -142,8 +143,12 @@ def rename_census_tract(self,row,row_num=None,column_name='census_tract'):
             column_name corresponds to the key of the row, which depends on 
             the source file column name which may be different from the final consistent name of census_tract
             '''
-            row[column_name] = self.census_mapping[row[column_name]]
-            return row
+            # deal with null values TODO - check with Neal what this is doing
+            if row[column_name] == 'Null':
```
NealHumphrey: should change `'Null'` to `self.null_value`

python/housinginsights/ingestion/insertion_functions.py
```diff
@@ -0,0 +1,299 @@
...
+
+def project_csv_fix_empty_address_id():
+    """
+    Identifies any rows in the project.csv file with missing 
+    'proj_address_id' and attempts to fix by using latitude/longitude, 
+    proj_x/y, or proj_addre/city/st/zip.
+    
+    :return: 
+    """
+    proj_df = pd.read_csv(os.path.join(DATA_RAW_PATH, 'preservation_catalog',
```
NealHumphrey: @jkwening So unfortunately we'll want to refactor this to not use Pandas by the time this is all done. Pandas forces us to get a bigger server. The only place we use Pandas currently is in the make_draft_json, which is only used in development (not production). 

If this is working we can skip that for now and deal with it later, but wanted to flag it for future development.

python/housinginsights/ingestion/insertion_functions.py
```diff
@@ -0,0 +1,299 @@
...
+                result = None
+
+        if result is not None:
+            proj_df.ix[idx, 'Proj_address_id'] = \
+                result['Table1'][0]["ADDRESS_ID"]
+
+    return proj_df
+
+
+def create_mar_csv(project_df):
```
NealHumphrey: We talked about switching this behavior to use the database and adding the appropriate column to the project table in the database. What do you think about when we should do that? Better to deploy using your csv file for now and adjust later? Or should we try to put in that refactor now?

python/housinginsights/ingestion/insertion_functions.py
```diff
@@ -0,0 +1,299 @@
...
+        file_name += '-' + time_stamp
+
+    path = os.path.join(path, file_name)
+
+    data_frame.to_csv(path, columns=data_frame.columns, index=False)
+
+    return os.path.isfile(path)
+
+
+def fill_empty_address_id(csv_file, address_id='Proj_address_id',
```
NealHumphrey: what is the difference between this and the project version?

python/housinginsights/ingestion/insertion_functions.py
```diff
@@ -0,0 +1,299 @@
+"""
```
NealHumphrey: Ultimately we should probably refactor these to be part of the relevant cleaners, so they fit more closely with the existing code structure

python/housinginsights/sources/base.py
```diff
@@ -164,7 +165,95 @@ def directly_to_file(self, data, filepath):
...
+                open(source_proj, mode='w', encoding='utf-8') as proj, \
+                open(source_subsidy, mode='w', encoding='utf-8') as subsidy:
+
+            reader = csv.DictReader(f)
+            proj_writer = csv.DictWriter(proj, fieldnames=PROJ_FIELDS)
+            proj_writer.writeheader()
+            subsidy_writer = csv.DictWriter(subsidy, fieldnames=SUBSIDY_FIELDS)
+            subsidy_writer.writeheader()
+
+            for line in reader:
```
NealHumphrey: Add some comments to show reason - makes sense to me but we've discussed how these things interact a lot, would be confusing to a new person.

"If the project doesn't exists in the database, we want to add a new record to the project table"

"All projects need a new record added to the subsidy table. If that project already exists in the database, link it to the project using the nlihc_id; if not, link it to the record that was added to the proj_writer output file."

python/housinginsights/sources/base.py
```diff
@@ -164,7 +165,95 @@ def directly_to_file(self, data, filepath):
...
+                                                  line=line)
+                    proj_writer.writerow(data)
+
+                data = self._map_data_for_row(nlihc_id=nlihc_id,
+                                              fields=SUBSIDY_FIELDS,
+                                              fields_map=subsidy_fields_map,
+                                              line=line)
+                subsidy_writer.writerow(data)
+
+        # remove source file so it doesn't get added to manifest hence database
```
NealHumphrey: Clarify why you'd want to remove the 'source' file. 

"source_csv is the data we are trying to merge with our existing project and subsidy tables. Since we have the data split into appropriate proj_ and subsidy_ csv files we no longer need the source_csv file."

Long term I might consider figuring out a way for the source_csv to be excluded from the merge but still not deleted so we can trace down problems if needed, but no big deal for now.

python/scripts/load_data.py
```diff
@@ -1,227 +0,0 @@
-"""
```
NealHumphrey: We should have something in the /scripts file that we can use to rebuild the database - essentially the same code you have after the if name=main statement in the /ingestion/load_data.py file.  That way we can keep the similar workflow where the /scripts folder is where people go to find executables if they don't need to know how the housinginsights package works.

python/housinginsights/ingestion/load_data.py
```diff
@@ -0,0 +1,458 @@
...
+sys.path.append(PYTHON_PATH)
+
+from housinginsights.tools import dbtools
+
+from housinginsights.ingestion import DataReader, ManifestReader
+from housinginsights.ingestion import CSVWriter, DataReader
+from housinginsights.ingestion import HISql, TableWritingError
+from housinginsights.ingestion import functions as ingestionfunctions
+
+# configuration: see /logs/example-logging.py for usage examples
```
NealHumphrey: Don't actually want this here - move it into the if name == main statement. If its here but the code is imported from another file (like from /scripts or from the API server) we'll end up with duplicated log statements (this comment from experience)

NealHumphrey: By this I mean the logging configuration statements line 43 through 49

jkwening:
@NealHumphrey I think this is now ready to go. I think I incorporated all your feedback. I left 'update_method' in the manifest but I ended up not needing it but I think this is useful information to track about the data sources. I may have hastily cleaned up the manifest.csv so you may want to make sure I didn't remove any rows that were still needed. Let me know if I missed something.

NealHumphrey:
@jkwening Looks good! I did a little bit of editing - reassigned some of the manifest.csv files to use the downloaded version, and added a load_data.py file back into /scripts with can be used to run the code. Note we will need to update the documentation. 

One thing missing is we need to load the meta.json into the database so that it is available from the api. I think this got put in between when you started this refactor and when we put this all together. I'll try to pull that in next. 

I pulled the latest data (except wmata_dist for the new dchousing records) from the apis and it is synced to S3 so when you run this now it should download those files for you.

PR #303, opened 2017-05-31
updates to filter

wassona:


NealHumphrey:
@wassona I merged this in, but went ahead and added the call to the API so I could test the code. Ended up making a few other changes to simplify how we know which filter options are available, instead of keeping an additional options log. The slider filter is working now; you should make sure to pull the latest dev branch from codefordc before working on getting the other data types running!

PR #310, opened 2017-06-05
273 jo building list filter and flyTo location

ostermanj:
This PR fleshes out some more of the responses to a filteredData state change, provides a list of buildings matching the filter criteria, connects actions on that list to map navigation.

* gives the right-side buildings pane a scrollable div of all  matching buildings
* list items animate a little on enter and exit to help understand what's happening
* the heading shows how many buildings match
* mousing over a list item highlights it on the map (after a half-second delay to avoid inadvertent highlighting)
* clicking on a list item flies the map to the project's location and opens its popUp for potential navigation to the building view
* provides a reset zoom button after zoom events to go back to original zoom settings
* legend is moved to be superimposed on the map, bottom center. (still hard-coded)

There's also styling changes to make the left and right sidebars more consistent with each other.

One issue found while working on this (I'll make an issue and link to it): the filters are not handling null values. 34 buildings, for example, have `null` project units counts and are filtered out when any filter is applied.


ostermanj:
filtering issue is #311

ostermanj:
Small conflicts with the latest merge from Paul's #298, but they're minor. Basically defer to Paul's conflicting changes in building-view.js and mine in map-view.js, with the exception that Paul's setState('selectedBuilding', building); will need to part on the popUp's click event.

ostermanj:
working now with latest dev, with building view refactor

docs/tool/js/views/map-view.js
```diff
@@ -396,37 +425,43 @@ var mapView = {
...
+            console.log(building);
             if ( building === undefined ) return;
-            setState('previewBuilding',{
-                proj_addre: building.properties.proj_addre,
-                proj_name: building.properties.proj_name,
-                hud_own_name: building.properties.hud_own_name
-            });
+            setState('previewBuilding', building);
+
+                /*{
```
NealHumphrey: Should this be removed? Or might we need to revert to this technique?

ostermanj: Thanks! Oh yeah -- the commented out code can be removed. I needed it for references, but no longer.

Issue #236, opened 2017-05-04
Produce list of filter criteria


Nunie123:
Being able to filter the result set of properties is a critical feature.  

In order to implement the filter feature, we need to decide on a list of criteria that the user will be able to filter on.  At this late stage of the project, we should limit our filter criteria to data that is currently ingested, or data that we expect to be ingested soon (consult with ingestion team).

Once we have decided on a list of criteria then the front end people can design and build the form, and the back end people can build the API.  This list will have a significant impact on the implementation and should be completed (at least preliminarily) before substantial effort is put into other parts of this feature.

Nunie123:
Potential filter criteria:

- Type of subsidy
- Agency providing subsidy
- Subsidy start date
- Subsidy end date
- At risk of leaving program
- Number of units covered by subsidy
- Active vs. former subsidized properties
- REAC score
- Last inspection
- Last renovation
- Avg. rent by neighborhood/ward/census tract
- Higher vs. lower crime by neighborhood/ward/census tract
- neighborhood/ward/census tracts with rapidly increasing rent
- Distance to closest rail/bus stop.
- Distance to a particular rail line.

What else can people think of?

NealHumphrey:
@Nunie123 - thanks for starting this! I suggest starting simple to get the ball rolling. The following fields are a) definitely available and b) definitely needed and c) represent a few different technical challenges for how the data is associated to a building. 

We will want more than this eventually, 

1) project -> 
This is the most simple filter, where one record means one project that is filtered in/out
- proj_units_tot
- proj_units_assist_max 
- hud_own_type
and 'geographical area':
- ward
- ANC
- geo2010 (will be renamed to census_tract)
- cluster_tr2000 (will be renamed to neighborhood_cluster)
- proj_zip (will be renamed to zip and current zip [which says things like "ZIP 20001"] will be named zip_string)

2) subsidy ->
project to subsidy is one to many. Want to return project records if ANY of the many subsidy records match criteria:
- poa_start  (this is subsidy start date, poa=period of affordability)
- poa_end
- portfolio (this is a higher-level 'type of subsidy')

3) census ->
Our 'acs_rent_median' table will soon be replaced instead with a 'census' table that includes records for ALL census fields (one of which will be acs_rent_median). When this is ready (after doing above) we'll want to add this as one example - but we will want to have many options. 

"in a census tract with...."
- acs_rent_median above/below the  median of all census tracts in the city. 


[...]

PR #318, opened 2017-06-07
Adds all filter variables requested in #236

NealHumphrey:
Also
- Alters the logic that builds the filter UI so that the list of variables is populated from the data available instead of hard-coded into the filterView.components
- Edits the API so that variables are no longer namespaced (this has been deployed to the API)

### DIFF
--- a/docs/tool/js/views/filter-view.js
+++ b/docs/tool/js/views/filter-view.js
@@ -10,14 +10,55 @@ var filterView = {
         //  the same meaning in both places (e.g. 'ward' and 'ward' can appear in two tables but should have same name/format)
 
         {   source: 'total_units_in_building',
-            display_name: 'Project unit count',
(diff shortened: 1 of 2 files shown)


NealHumphrey:
@ostermanj @ptgott @Nunie123 and @HabRonan 
Heads up - I have added in all the 12 filter variables, as well as made the categorical variables dynamically update. 

You need to merge this (the current dev branch) into your current code because it also involves an API update. Previously the API was namespacing some of the column names of the filter table to more descriptive names. I undid this so now the names match exactly what they are in the database - this was needed to play well with our meta.json auto documentation stuff. If we want to change a column name to something else, we can certainly do it but it needs to be done through the whole workflow (data ingestion to api to front end). 

I deployed this API change to the live servers so that I could test it locally; therefore I went ahead and approved my own merge request and merged this into dev.

PR #319, opened 2017-06-07
Ready - Transforming layer menu to V1 launch version.

NealHumphrey:
- Instead of a simple `<a>` link, adds full set of divs including title and content for each layer option
- Added semantic-ui accordion styling to open close
- rewrote callback for toggling data layer based on user selection.

Still to do:
- move up the call stack and make code work with an edited configuration json file. Move it out of our temporary meta.json and into its own config file. This will allow handling custom data endpoints (i.e. not just a straight count of a whole table)
- Add initial list of data layers for V1.


NealHumphrey:
My most recent commit does lots of stuff:

Converts our approach to overlays to enable expanding to more layer choices
- Shifts logic for choosing which layer choices to show out of meta.json and into a separate configuration json, currently hardcoded into mapView.buildOverlayOptions() (will shift out later)
- Adds descriptive field for each data source
- Enables persistence of selection for layers, so whether you select a new zone or a new data layer the choice of the other one (zone or layer) is persisted. 
- Implements 'default' zone types that are triggered when the user 
- Handles the case of 'null' data returned from a getData request. For all current cases, the code has been set up to handle a state of 'dataLoaded = false' just by logging errors. Typically this only occurs currently for the case of selecting an aggregation type not available for a particular data set. 

The interaction of 'layer' (i.e. zone boundaries) and 'overlay' (i.e. chloropleth) had some complicated looping logic - they are treated as separate layers and behave independently, so for example if no data is selected you can still toggle the zone boundaries. The logic for the different things that need to happen when one is clicked but not the other are a bit messy. Unfortunately I don't think I cleaned it up a lot - it's still a little complicated - but for current purposes it is now behaving as desired. 

The logic of 'previous layer' vs 'current layer' was causing the most problems, and is not totally cleaned up. Since we are forcing the user to only show one layer at a time (instead of when we showed how boundaries overlapped in rich-map.js), for the 'layers' section I made it simpler by always removing all existing layer choices before adding in the new one. It would likely be cleanest for our 'overlay' logic to adopt that approach instead of always trying to figure out what the previous layer was and picking that off to hide. This will enable cleaner 'external' triggering of layers on or off, i.e. if we add any other way to do this besides the current toggles. 

Still to do: disable buttons for zone types that aren't available

NealHumphrey:
Got another commit, got a bunch more stuff. This is all I'll do for now - I'll have a few more hours tomorrow too but will mostly be away from computers this weekend. If anyone has time for a quick review to at least test functionality that'd be good, but regardless to keep things moving I suggest merging this into dev before you do other work - this PR touches most of the front end code. 
@ostermanj or @ptgott - if either of you work on this tonight


- Adds accordion panels to the Filter page
- Fixes the callback on the categorical selectors that was being attached to all 'select' elements instead of just the specific one that was triggered
- Mostly fixes the logic issue where nulls were permanently excluded. Sets the value of the array to `[]` when sliders are set equal to min and max (i.e. back in their original state), which is the same logic as categorical (which also gets an empty list when none are selected, which means 'all' instead of 'none').
- Adds color coding to the accordion panels on Filter to change color when that filter is activated, and to change back when the filter is cleared.
- Adds descriptive text to both filter and layer data elements

PR #320, opened 2017-06-07
Generate data required in issue 140

HLKwon:
Regarding issue 140.
@NealHumphrey @jkwening

DONE (@ilissablechwapo @HLKwon @ncclementi):
- Able to generate CSV file with total housing and bedroom units in a specific zone (ie. ward, zip, anc, census, etc.).
- csv file is created in `data/processed/zoneUnitCount`.
- the current python file that creates the csv file runs like a script, `cama.py` located in `python/housinginsights/sources`.
- Was able to ingest the csv file to create the table in the local DB by following the instructions in the [load data tutorial](http://housinginsights.org/resources/adding-data-sources.html).

TO DO:
- Will refactor the code in the coming weeks in order to be able to run in `load_data.py` (@HLKwon @ncclementi)

Will request to close the issue once the code is refactored.


NealHumphrey:
Hi @HLKwon , @ncclementi @ilissablechwapo - thanks for your work on this! 

I'll review this weekend - sorry, been focused on front end stuff the past couple daysy. It would be great to include this data set in V1. As noted in your 'TODO', we definitely want to refactor it so that it works with load_data.py. But, assuming I don't see any logic issues I will probably merge into 'dev' even if it does not yet do this; that way we can include the data in V1 and make sure that it is ready to work with load_data.py (and thus be able to be updated automatically) for V2.

NewFile
```diff
@@ -0,0 +1,8 @@
+web:
```
NealHumphrey: I don't think this file is used/needed. Can you remove from the pr?

docker-compose.yml.back
```diff
@@ -0,0 +1,22 @@
+version: '2'
```
NealHumphrey: What is this file for? Remove if not needed.

python/housinginsights/secrets.json.back
```diff
@@ -0,0 +1,37 @@
+{
```
NealHumphrey: @HLKwon definitely need to remove this - I'll need to reset all these keys now. Was planning to anyways as part of the V1 launch as we shift over to the Code for DC account, but be sure to never commit secret keys like these ones.

python/scripts/fetch_DC_housing_unit_data.py
```diff
@@ -0,0 +1,8 @@
+import csv
```
NealHumphrey: Should this file be removed? It looks like it was just test/demo code.

python/housinginsights/sources/cama.py
```diff
@@ -0,0 +1,208 @@
...
+import datetime
+
+PYTHON_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir, os.pardir))
+sys.path.append(PYTHON_PATH)
+
+from housinginsights.sources.base import BaseApiConn
+from housinginsights.sources.models.mar import MarResult, FIELDS
+
+
+class MarApiConn(BaseApiConn):
```
NealHumphrey: Note that we already have a MarApiConn class in the sources/mar.py file . I suggest renaming this one for now (e.g. SecondMarApiConn), and during your refactor see if you can get it to just use one single MarApiConn class.

python/housinginsights/sources/cama.py
```diff
@@ -0,0 +1,208 @@
...
+                           'neighborhood_cluster': entry['CLUSTER_'],
+                           'ward': entry['WARD'],
+                           'zip': entry['ZIPCODE']
+            }
+
+        return mar_returns
+
+class CamaApiConn(BaseApiConn):
+    """
+    API Interface to the Master Address Record (MAR) database.
```
NealHumphrey: This docstring needs to edited, looks like it was copy/pasted. Would be good to be more descriptive - e.g. what is Cama, what kind of data do we retrieve from cama

python/housinginsights/sources/cama.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+        """
+        Take each CAMA property data and retrieve the MAR data.
+        """
+        """
+        Certain square values have four digits + a letter. (ex. 8888E)
+        Square would be the first four digits and suffix would be the letter.
+        SSL sometimes comes as 8 digit string without spacing in the middle.
+        """
```
NealHumphrey: Thanks for includeing this documentation, very useful!

python/housinginsights/sources/cama.py
```diff
@@ -0,0 +1,208 @@
...
+                if len(square) > 4:
+                    square = square[:4]
+                    suffix = square[-1]
+
+                mar_return = mar_api.get_data(square, lot, suffix)
+
+                ''' Count the housing units and bedrooms '''
+                num_units = 0
+                if row['properties']['NUM_UNITS']: num_units = row['properties']['NUM_UNITS']
+                if num_units == 0:
```
NealHumphrey: Do you have any documentation that says that this is a valid assumption? Is 0 for studios or something?

python/housinginsights/sources/cama.py
```diff
@@ -0,0 +1,208 @@
...
+                    num_units = 1
+
+                bedrm = row['properties']['BEDRM']
+                if bedrm == 0: bedrm = 1
```
NealHumphrey: Again, wondering about assumptions here

python/housinginsights/sources/cama.py
```diff
@@ -0,0 +1,208 @@
...
+                        if not flag:
+                            zone_count.append( OrderedDict([('zone_type', zone), ('zone', mar_return[zone]), ('housing_unit_count', num_units), ('bedroom_unit_count', bedrm)]) )
+
+            except Exception as e:
+                exc_type, exc_obj, exc_tb = sys.exc_info()
+                print(exc_type, "line", exc_tb.tb_lineno)
+                print("Error! SSL: ", row['properties']['SSL'], row['properties']['AYB'])
+                continue
+
+        return {'anc': anc_count, 'census_tract': census_count, 'neighborhood_cluster': cluster_count, 'ward': ward_count, 'zip': zipcode_count}
```
NealHumphrey: During the refactor, to get this to behave like other ApiConn objects, the 'get_data' should be the one that produces the csv.

python/housinginsights/sources/cama.py
```diff
@@ -0,0 +1,208 @@
...
+
+        keys = toCSV[0].keys()
+
+        with open(filename, 'w') as output_file:
+            dict_writer = csv.DictWriter(output_file, keys)
+            dict_writer.writeheader()
+            dict_writer.writerows(toCSV)
+
+
+my_api = CamaApiConn()
```
NealHumphrey: To make this able to be imported into /python/scripts/load_data.py, all you have to do is add an `if __name__ == '__main__':` statement just above this line. that way it won't automatically run when it is imported, instead it will make the objects available; but it will run if you run this specific file from the command line.

ncclementi:
@NealHumphrey @HLKwon I figured out how to push to Louis fork, I'm deleting this files that I assume were generated while testing the docker locally. Same with the other files that look they weren't needed. I assume that all these files were added by mistake because of the use of `git add .`

python/housinginsights/sources/cama.py
ncclementi: Renamed to `MarApiConn_2` and yes the idea is to add what is there as a different method in the original one and then import the class.

python/housinginsights/sources/cama.py
ncclementi: @HLKwon You had an explanation for this, what was it?

python/housinginsights/sources/cama.py
ncclementi: Same here, @HLKwon You had an explanation for this, what was it?

python/housinginsights/sources/cama.py
ncclementi: Got it We'll have this in mind!

python/housinginsights/sources/cama.py
HLKwon: In the CAMA data, buildings with 0 num_units (the data for number of housing units) falls under number of cases. There are about 177 data points with 0 housing units. 

- First set of CAMA data that has 0 num_units and 0 bedrooms. However, they have associated MAR data points. Also their use codes are residential. Some examples are,
    (SSL: 1264 0808) // according to Integrated Tax System API: land use code 023 - Residential Flats-Less than 5
    (SSL: 1256 0078) // land use code 012 - Residential-Detached-Single-Family
The assumption was that these buildings are used for the purpose of residential because they are taxed as such. Therefor there should be at least 1 bedroom and therefore it counts as 1 housing unit. There should be less than 10 data points that meet this criteria, so even if the assumption is off, it shouldn't really be that big of a deal in a data point with 100,000+ data points. 

- Second set is certain data points for single family homes (and the most numerous data points with 0 num_units). The have multiple bedrooms but they show up as 0 num_units even though it should be 1. They also have MAR data point when you call it on the MAR API. An example is (SSL: 0069 0805). CAMA data shows 0 num_units and 3 bedrooms. MAR API has appropriate data. When looking at Integrated Tax System API you see building code 011 for Residential-Row-Single-Family. These housing units should be counted as 1 housing unit instead of 0, as they contain bedrooms and serve as a housing unit. There are about 154 data points out of 177.

- Third set of data point with 0 housing units are data point with 0 housing units and 0 bedroom but has no MAR data associated with them. An example is (SSL: 0514 0871). They are excluded from our data points as there is no way to figure out which zone they belong in. There are less than 20 data points are in CAMA. 

Summary: There are data points in CAMA with 0 housing units, but most are single family or equivalent housing stock and should be counted as 1 housing unit instead of 0.

python/housinginsights/sources/cama.py
HLKwon: In the CAMA data, there are about 207 data points that have 0 bedrooms. 
- 172 out of 207 have 1 housing unit but contains "0 bedroom". Many has land use code of 011 - residential single family housing. Assumption was made that 1 housing unit should be equivalent to 1 bedroom. 
- There are also data points like above with 0 housing units and 0 bedrooms. I made the assumption that if the MAR data exists then there should be at least 1 housing unit and 1 bedroom.

python/housinginsights/sources/cama.py
NealHumphrey: @HLKwon Great, thanks for the information! 

Can you add a new file to `/docs/data/cama.md` where we will add documentation for this data set, and then add these comments into that file? We'll want to have this written up in our public-facing documentation in case anyone has questions. 

In the code, this can then just have a quick note that says 'see docs/data/cama.md for information on assumptions`

NealHumphrey: Note this can be in a separate pull request

PR #322, opened 2017-06-08
Jo connect charts to filters

ostermanj:
— Two donut charts instead of four
— Connected to filterData event
— First shows percentage of buildings matching the filter(s)
— Second shows the percentage of total units represented by the buildings matching the filter(s)

The last commit puts the charts in the buildings pane and removes the charts pane. With only two charts, it seems unnecessary to have a separate pane for them. And these charts work really well here, animating their change at the same time the list is shortened or lengthened.

If you prefer keeping the charts pain, revert to the second to last commit.

### DIFF
--- a/docs/tool/css/styles.css
+++ b/docs/tool/css/styles.css
@@ -47,6 +47,7 @@ html, body {
     margin-top: 40px;  
     display: none;
     overflow-y: auto;
+    padding-top: 10px;
     
 }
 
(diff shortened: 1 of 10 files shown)


NealHumphrey:
Awesome stuff! I agree, it's good to have the charts above the building list. 

Previously we talked about having summary charts related to both the building and overlay data. This could be a better solution to that - having a couple building charts (like these two, maybe add a couple more in V2) above a half-height building list, and use the second tab exclusively for overlay related data. Of course we still need to decide what summary charts would even make sense to have related to our layers. We can consider putting the second pane back for V2 - and could even link the toggling of the right pane to the toggling of the left pane so that you only have one choice to make (buildings or layers) and the unified view switches out.

Issue #313, opened 2017-06-06
Cache walking distance calculations


NealHumphrey:
The walking distance calculation is very time consuming (4+ hours) due to the rate limit imposed by Mapbox. In reality we only need to run this once per pair of locations (building+station). Stations and buildings rarely move (mostly in cases of data errors not actually due to moving). However, if the lat/lon of a building or station changes (though the ID of that station did not change) we do want to rerun the calculation. 

Here's a proposed quick solution that has minimal impact on the rest of our workflow as it doesn't require us to do tricky state management of changing lat/lons for specific ids. 

- Add the 4 numbers for start and end lat/lon to the wmata_dist table, to cache the specific lat/lon values that were pulled from the project and wmata_info tables when the distance was calculated. 
- Add an optional argument to WmataApiConn.__init__ for 'use_cached_distance' and set it to True by default. if self.use_cached_distance is true, we don't need to get the walking distance if we already have it. if False, we'll behave as normal and get fresh walking calculations for all points. 
- Edit the `self._get_walking_distance` method to first query the database for a row corresponding to that exact set of 4 lat/lon values. If the wmata_dist table does not exist (i.e. we are building it the first time or want fresh data) or if there is no matching row (either because it's a new building or because one of the lat/lon values has been changed) proceed with getting the walking distance from the API.

NealHumphrey:
@jgordo04 If you're not too sick of working on distance calculations - this and #312 are two enhancements so that our server won't have to run the wmata_dist code for 5 hours every week!

jgordo04:
No prob. Happy to!

jkwening:
Cleaning up related data ingestion issues - closing because it looks like this has been completed.

Issue #312, opened 2017-06-06
Add crow distance to wmata_dist table


NealHumphrey:
To speed up processing of rail distances, I pulled in the haversine calculation of crow distances based on lat/lon pairs. Since we have this calculation available, we might as well add it to the database. 

Currently the calculation is only in the _find_rail_stations method. 
- Add calculation to the _find_bus_stations method
- Add it to the distOutput as a new key value pair ('crow_dist': 0.5)
- Add new column to meta.json
- Run the process from start to finish (download data, update manifest, add to database) to make sure it all works. 

Not urgent, plan for v2

jkwening:
Cleaning up related data ingestion issues - closing because it looks like this has been completed.

PR #323, opened 2017-06-10
Changes needed: Added use_cached_distance and database checks for walking distance for issues #313 and #312

jgordo04:
Added changes for issue #313

### DIFF
--- a/python/housinginsights/sources/wmata_distcalc.py
+++ b/python/housinginsights/sources/wmata_distcalc.py
@@ -246,11 +246,14 @@ class WmataApiConn(BaseApiConn):
                 if ( 'building_lat' in columns and 'building_lon' in columns and 'stop_or_station_lat' in columns and 'stop_or_station_lon' in columns ):
                 #See if row exists
                     proj_query = 'select * from wmata_dist where building_lat=\'' + srcLat + '\' and building_lon=\'' + srcLon + '\' and stop_or_station_lat=\''+ destLat + '\' and stop_or_station_lon=\'' + destLon + '\''
-                    rows = conn.execute(proj_query)
-                    if ( rows != None ):
+                    row = conn.execute(proj_query)
+                    if ( row != None ):
                         conn.close()
                         engine.dispose()
-                        return None
+                        result = []
+                        for r in row:
+                            result.append(r)
+                        return result
 
                 conn.close()
                 engine.dispose()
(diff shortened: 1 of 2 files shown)


jgordo04:
All of #312 is done as well, except for running the process from start to finish

jgordo04:
The changes to walking_dist basically return nothing if it already exists, assuming we will append to the current database. I can change it to return the already calculated dist if that is preferred.

jgordo04:
The changes repopulate the csv file with the database results if they already existed. If you want them not in the results at all, just let me know and I will change it

python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -275,15 +309,19 @@ def _find_rail_stations(self, railStations,project_details,radiusinmeters,sample
 
         for idx, station in enumerate(railStations):
             crow_distance = self._haversine(lat, lon, station['Lat'], station['Lon'])
-            
+
             if crow_distance < (radiusinmeters / self.meters_per_mile):
-                walkDist = self._get_walking_distance(lat, lon, str(station['Lat']), str(station['Lon']))
-                walkDistMiles = walkDist / self.meters_per_mile
+                walkDist = self._get_walking_distance(lat, lon, str(station['Lat']), str(station['Lon']),db=db)
+                if type(walkDist) is not list:
```
NealHumphrey: I'd recommend not having `self._get_walking_distance` return two different types of data in two different circumstances (cached or not). Instead, I'd recommend have _get_walking_distance always return a distance, and then this code can just  use the same append statment in line 321. This will make it more maintanable / easier to understand for future developers.

python/housinginsights/sources/wmata_distcalc.py
```diff
@@ -228,6 +229,38 @@ def _get_walking_distance(self, srcLat, srcLon, destLat, destLon):
...
+                #See if row exists
+                    proj_query = 'select * from wmata_dist where building_lat=\'' + srcLat + '\' and building_lon=\'' + srcLon + '\' and stop_or_station_lat=\''+ destLat + '\' and stop_or_station_lon=\'' + destLon + '\''
+                    row = conn.execute(proj_query)
+                    if ( row != None ):
+                        conn.close()
+                        engine.dispose()
+                        result = []
+                        for r in row:
+                            result.append(r)
+                        return result
```
NealHumphrey: See my comment below, I recommend switching this so that only the distance is returned so that _get_walking_distance behaves the same way in both cases.

jgordo04:
@NealHumphrey , changes made!

NealHumphrey:
Cool! Testing now

Issue #317, opened 2017-06-07
Front-end home-stretch items for June 13


ptgott:
Based on a conversation at yesterday's meeting, here is a condensed version of the to-do list in Issue #316. The list reflects the number of front-enders who'll likely have time to work on this stuff through the next week.

(EDIT by Neal: I got a lot done so I pulled over the whole list and updated with status as of 6/8. Ones marked 'John' are ones he put his initials on in #316, but he's out of town Friday onward so can reclaim those if needed)

### Do first: Things we definitely need by June 13
- [x] Add the remaining filters to the Filter panel.
- [x] **Polish the Filter and Layer panels**: unifying color schemes (blue means context, orange means building) and adding accordions to fit all of the options in each panel. This can also include unifying the visual elements of the Building View with those of the Map View.
- [x] (Paul) **"Clear all filters" button** that appears at the top of the map panel when at least one of the filters has been applied.
- [x] **Rather than use 'meta.json' to auto-populate options** for selecting data layers, aggregate zones and (possibly?) filter fields, use a separate .json file. (Note: logic step is complete. Currently this json is hard-coded into our code - can put off moving into a separate file until V2)

### Big list of all the other stuff moved here from #316
Browse:
- [x] (Neal) Finalize a list of chloropleth data sources (will involve adapting how the data sets are chosen since there will often be more than one per table and they will sometimes be named something other than the table name, e.g. 'Building permits in the last 3 months' or '% change violent crime in past 5 years')
- [x] Add a div wrapper around the data type link and add a ? button that links to our data documentation. (note: added short description but still need to link to docs)
- [x] Implement persistent selection of zone type when selecting a new browse data source
- [x] Disable buttons when aggregation method is not available
- [ ] Style the Layers panel in blue theme to link data to controls (partially implemented)
- [x] (Paul) **Add a legend** for the chloropleth values

Filter:
- [x] Add all 12 initial data sources proposed. OK to hard-code the meta data (filterView.components) for now rather than building out. 
- [x] Style w/ orange color palette (partially done)
- Pillboxes w/ 'x' to clear filter
[...]

PR #324, opened 2017-06-10
317 clear all filters

ptgott:
This has to do with issue #317. There are a few things to note:
1. The 'clear filters' button clears the filters by calling a 'clear' method on each filter-related UI component, which I've extracted into a constructor. This is because the control of the filter values is tied to the state of the UI components. This means that resetting the UI components and resetting the filtered values is one and the same move.
2. There's already a 'reset zoom' button in the center-top of the map view, and I didn't want to interfere with this. As a result I've based the Clear Filters button on the deletable labels that appear in the SemanticJS dropdown menu, adding the option to clear the filters to the filter tab itself. If there's a more intuitive option, I'm okay with us changing it.
3. For some reason, the sidebar toggles whenever the user clicks on the Clear Filters button. This has something to do with the sub-nav-button's 'click' event triggering/propagating, though I'm not sure why.

### DIFF
--- a/docs/tool/css/filters.css
+++ b/docs/tool/css/filters.css
@@ -58,5 +58,10 @@
   left: 10px;
 }
 
+/* override the Semantic.js label styling for the Clear Filters button */
+#clearFiltersPillbox{
+    font-size: 1em !important;
+}
+
 
 /**/
--- a/docs/tool/js/views/filter-view.js
+++ b/docs/tool/js/views/filter-view.js
@@ -98,9 +98,6 @@ var filterView = {
             min: '1950-01-01', //just example, TODO change to date format
             max: 'now'         //dummy example
         },
-
-
-
        
     ],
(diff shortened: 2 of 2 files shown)


NealHumphrey:
Thanks @ptgott ! 
1. I think that your approach to refactoring into a constructor with a single 'clear' method is a good move. 

I reviewed and functionality worked, and most of the approach looks good. But, there's one logic issue about how filtering/clearing events are triggered that is causing some minor performance issue that will become major as we add more filters. I decided to make a separate issue for it instead of delaying merging: #333 . 

2. I think it will be better to not mix navigation and clearing. But I agree with mimicking the clearable labels. Today I will be doing some other cleanup adding subtitles, stuff like that, so can move the button around and see if there's a more natural home for it. 

3. I think this will be moot when we move the button off the navbar.

Issue #316, opened 2017-06-06
Home Stretch - Final Items for June 13


NealHumphrey:
For discussion and editing in our session tonight - my first pass at what I see remaining. TBD how many of these we can complete. 

Data ingestion:
- [x] @jkwening Add meta.json back into the API
- [x] @jkwening Automatic update of the manifest.csv when the get_api_data is run
- [x] @louvis (and/or @HLKwon @ncclementi) Ensure consistent column name and data structures for all current tables
-  [x] @MikeS-nth  Add ~15-20 census data fields
-  [x] @ajhalani Add weighting factors for census data?
- [x] @emkap01 (+others) Data documentation pages for all data sources we actively have (including JS to pull in field descriptions from the meta.json from the API)

API
- Add endpoint for census data that applies weighting factors?
- Add endpoint for Filter categories / configuration? Or build it on the front end
- Verify that for chloropleth/browse data sources that we know which aggregation methods we can use (i.e. neighborhood_cluster, zip, etc.) - with consistent column name / data structure implemented our existing endpoint should work. 

(edit: moved all front end items to #317)

emkap01:
Data documentation for all raw datasets used in the production db is being worked on, the status of each one is being tracked via [this sheet](https://docs.google.com/spreadsheets/d/1EcjViJq2SuiHEEiRu9TemVJ29oNrNV7ibC6NofJiXzA/edit?usp=sharing).

PR #325, opened 2017-06-11
Issue #316 - documentation for Subsidy and TOPA tables

emkap01:
This PR contains Markdown documentation files for both the 'Subsidy' table and the 'TOPA' table, which partially address the 'Data ingestion' section of Issue #316 .

@NealHumphrey I will defer to you to coordinate with the Filter team if need be to make sure that the JS for the front end picks up these two new files.

There should be a few more documentation files coming from @speedyturkey and @kasao10, per the spreadsheet we set up to divvy up the work: https://docs.google.com/spreadsheets/d/1EcjViJq2SuiHEEiRu9TemVJ29oNrNV7ibC6NofJiXzA/edit#gid=1802468479.

I thought we had accounted for all the datasets that still need documentation in there, but it looks like we may need someone to work on a Markdown file for the census data.  I will do a little more digging and pick that one up myself if need be.


emkap01:
This PR is a mess, I only wanted to include the two most recent commits.  I will close it and submit a cleaner one.

PR #327, opened 2017-06-11
PR for Issue 316 - documentation for subsidy and TOPA tables

emkap01:
This PR provides some of the documentation outlined via Issue #316. 

I am working on documentation for the 'census' table and will submit that when it's ready (might not get to it until tomorrow or Tuesday though).

### DIFF
--- a/docs/data/subsidy.md
+++ b/docs/data/subsidy.md
@@ -1,0 +1,29 @@
+---
+layout: http://www.neighborhoodinfodc.org/dcpreservationcatalog/
+tablename: subsidy
+title: Subsidy
+---
+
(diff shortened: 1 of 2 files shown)


docs/data/subsidy.md
```diff
@@ -0,0 +1,29 @@
+---
+layout: http://www.neighborhoodinfodc.org/dcpreservationcatalog/
```
NealHumphrey: @emkap01 'layout' is a specific variable used by Jekyll to tell it which format to use for the page. In this case `datasource` is the correct value.

docs/data/topa.md
```diff
@@ -0,0 +1,14 @@
+---
+layout: Parsed from DCHD data by Urban Institute staff
```
NealHumphrey: ditto on layout

PR #332, opened 2017-06-12
Added styling to buildings list side-bar.

Nunie123:
My apologies for the large number of listed changes.  A package on my text editor cleans up spacing in files I save, but plays hell with github's diffing (I've finally disabled the package).

The only lines I substantively changed are:

- lines 642-675 in map-view.js, and
- lines 214-224 in styles.css

These were just minor changes to makes the right panel a little bit nicer looking.

### DIFF
--- a/docs/tool/css/styles.css
+++ b/docs/tool/css/styles.css
@@ -4,33 +4,30 @@
     box-sizing: border-box;
 }
 
-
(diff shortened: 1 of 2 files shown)


NealHumphrey:
Sounds good, merging

PR #339, opened 2017-06-13
132 combo replace append

jkwening:
Completed what the following we discussed via email:

- Make a new file in /ingestion called Manifest.py
- Move the ManifestReader out of the DataReader.py file, and rename it just 'class Manifest()' (since it will now be doing both reading and writing). It can still inherit from HIReader and everything in it now can behave the same.  I think the only place you'll need to rename this where it's used is in your LoadData class. 
- Add a method to the Manifest class that takes a folder (i.e. the datestamped folder from the API run) and updates the manifest.csv in place

Added call to the get_api_data.py file that instantiates the Manifest class and then runs its update_api_data method.

TODO: At that point we can write the module that automates the process as you suggest (get api, update manifest, updates database)


jkwening:
Now also includes the fix for missing meta.json in database.

python/housinginsights/ingestion/Manifest.py
```diff
@@ -0,0 +1,135 @@
...
+        # update the manifest.csv in place
+        for row in self:
+            row_uid = row['unique_data_id']
+            if row_uid in uid_list:
+                row['data_date'] = data_date
+                filepath = os.path.join(file_path_base,
+                                        "{}.csv".format(row_uid))
+                row['filepath'] = filepath
+            else:
+                # parse the date back into format used in manifest
```
NealHumphrey: @jkwening can you elaborate? If we change the hardcoded values to YYYY-MM-DD can we remove this section and just do data.append(row) directly?

I think it actually would work this way regardless if we just remove this else block - as soon as we run it the first time, the manifest.csv would then have the year-first date formats, but the `__iter__` method dateparser I think is generic enough to handle either way?

The reason the date format is mm/dd/yyyy is because Excel automatically puts them that way, and I'm usually using Excel to manually edit the manifest. So `__iter__` needs to be able to handle either way in case the file is opened and accidentally saved with month first format, but when auto-writing year first format can certainly be default

NealHumphrey:
Note that I just committed a change to ignore your version of manifest.csv and instead use dev's, as it has some file changes. Let me know if there was any changes to your manifest.csv that you were expecting should carry through.

python/housinginsights/ingestion/Manifest.py
```diff
@@ -0,0 +1,135 @@
...
+                # TODO - manifest
+                try:
+                    _date = dateparser.parse(row['data_date'], dayfirst=False,
+                                             yearfirst=False)
+                    row['data_date'] = datetime.strftime(_date, '%m/%d/%Y')
+                except ValueError:
+                    row['data_date'] = 'Null'
+            data.append(row)
+
+        with open(self.path, mode='w', encoding='utf-8') as f:
```
NealHumphrey: Note, when I ran this it produced extra lines. We've encountered this in other places in the code too. The solution is to open with `newline=''` i.e. 
`with open(self.path, mode='w', encoding='utf-8', newline='') as f:`

I've made this change and will push it to dev when finished testing locally.

python/housinginsights/ingestion/Manifest.py
NealHumphrey: Verified that this works. I'm switching the default format of the manifest to be year-first, but the `__iter__` method used by LoadData will still work even if it is mixed month-first and year-first formats.

PR #340, opened 2017-06-13
Fix meta in db

jkwening:
Fixed the issue with missing meta table in the db. Now when the database is rebuilt, the meta table is also added.


jkwening:
@NealHumphrey going to close this pull request and just include this fix into the Manifest PR

PR #343, opened 2017-06-14
Fixes layer toggles

NealHumphrey:
Deploying to live site - full-stack changes to get the Layers to toggle back and forth correctly, including disabling of zone types that are not supported by a specific layer. 

Also includes John's updates for the LoadData that allows auto-updating of the manifest.csv


NealHumphrey:
@ostermanj and @ptgott - I'm merging as is, I've tested locally, and this stuff on our live site is pretty buggy so even if I missed something it's an improvement. This cleans up the layer/overlay logic a little bit, but I think it's worth considering a more thorough overhaul of that.

PR #346, opened 2017-06-15
180 census weighting plus crime count filtering

NealHumphrey:
This both adds the census weighting factor and expands our ability to do slicing and dicing of any 'observation' based table (i.e. where each record is some sort of unique item, like building_permits or crime). 

- Fixes bug where census_tract 1 was being ommitted due to Cleaner expecting the manually downloaded file format as opposed to currently-used API download format. Manually downloaded file had 2 header rows, api version has 1
- Adds flexible function for `census_with_weighting` that returns the weighted count of any census field in the 'items' format that can be sent straight to the json output
- Replaces the 'count_all' endpoint with a 'summarize_observations' endpoint. This one has significantly more flexibility including filtering of records (e.g. only violent crime or only construction building permits) and the ability to specify the desired time period.
- Adds method to divide two sets of itemized data, and to scale values by a factor.
- Adds the remaining 5 calculated census-based fields and updates poverty_rate to use the census_weighting approach so that ward and neighborhood_cluster are also available. 

Note that none of our API, including these updates, are particularly error or malicious proof - bad inputs create stack dumps and there's no SQL injection protection. That's on the list for V2

@ajhalani your review on this would be very appreciated!
@jasonrhaas if you're able to do a review as well it would be helpful. The challenges of applying weighting factors and adding flexibility for various types of ways of slicing our observation tables add some new things to consider for long-term approach and refactor of the API. 

@jkwening and @Nunie123 heads up

I will probably be going ahead and deploying this as I've tested moderately locally and want to get this data into the user testing so need to deploy before I leave Saturday.

### DIFF
--- a/python/api/application.py
+++ b/python/api/application.py
@@ -5,6 +5,9 @@ flask api
 ~~~~~~~~~
 This is a simple Flask applicationlication that creates SQL query endpoints.
 
+
+TODO, as of 6/15/2017 none of these endpoints are SQL Injection ready
+
 """
 
 from flask import Flask, request, Response, abort, json
@@ -27,6 +30,7 @@ from flask import jsonify
 from flask.json import JSONEncoder
 import calendar
 from datetime import datetime, date
+import dateutil.parser as dateparser
 
 #######################
 # Setup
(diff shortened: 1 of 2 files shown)


ajhalani:
I know this is merged for upcoming demo, but will try to review this weekend if there are any ideas for future enhancements.

NealHumphrey:
@ajhalani great thanks! I did already make one fix today to deal with situations where missing items are returned, which I'm merging in w/ the front-end code that fixed it. Basically whenever you review it's probably best to take a look at the diff of application.py from before this PR to the most recent dev.

Issue #344, opened 2017-06-14
Quick fixes this week for V1


NealHumphrey:
Pulling the top priority items out of [this google doc](https://docs.google.com/document/d/1R5nEh86T5yj1W4N3eqgiHHzZx9-QgTG4UxL8c76s2-k/edit)

- [x] Shrink circle size at moderate zoom level to avoid overlap
- [x] Add pillboxes for each filter that's selected along the top of map, as in mockup (this method provides better feedback as to what is currently filtered when you switch over to the layer view)
- [ ] Add a 'more info' link to each data explanation linking to the appropriate  table in our docs (probably want to do it via an additional configuration key in the initialOverlays variable of map view, and edit the overlayMenuOption function to append the appropriate link)
- [x] Move building dots to be on top of the blue layer so that coloring is not so washed out (note, there may be a few lingering logic places that rely on layer order, so there might be some logic fixes needed)
- [x] Move location of zoom + / - control so that it is not hidden by filter/layer panel. Bonus if it can move dynamically on filter open/close
- [ ] ~~Only animate entering/exiting buildings~~ (made new issue for this, not simple to implement)
- [x] reduce the size of dot swell animation since we now also want people to notice the animating pie chart. 
- [x] Back button more prominent in Building view. 
- [x] Rename “Buildings (#)” to “Results: # Buildings” 
- [x] Header of Buildings tab that describing donut charts:  “Your filters match 121 out of 545 buildings. This is: " and change "Buildings" to "of all buildings" and Units to "of all units"
- [x] Fix ACS median rent filter. Currently every item in the filterData collection has an 'acs_median_rent' of 'null'. As of 17 June, the acs_rent_median API endpoint doesn't seem to be working ([see](http://hiapidemo.us-east-1.elasticbeanstalk.com/api/raw/acs_rent_median)). 
- [x] Add more data sources to layers

API:
- [x] Update the census endpoint to apply weighting factors to create Ward and Neighborhood Cluster data correctly
- [x] Add 5 remaining calculated factors to census endpoint

NealHumphrey:
I've got most of the pillboxes (item 2) working - got them showing up / disappearing, just need to style slightly and add the callback function

ptgott:
@NealHumphrey I've updated the line item related to acs_rent_median. This seems to be a back-end issue, as all of the values received from the filterData endpoint are 'null'.

ptgott:
@NealHumphrey I'll be working on some of the quick fixes in #355.

[...]

PR #348, opened 2017-06-16
344 pillboxes

NealHumphrey:
Adds the pillboxes. Here's a demo:
demo_pillboxes

@ptgott or @ostermanj a quick review would be helpful! I'll be merging by end of day today so it'll be ready for demos before I leave, but a post-merge review to check for bugs still welcome.

### DIFF
--- a/docs/tool/css/styles.css
+++ b/docs/tool/css/styles.css
@@ -82,6 +82,14 @@ html, body {
 #options-spacer {
     flex-grow: 1;
     border-bottom: 2px solid rgb(189, 54, 33);
+    position: relative; /*so pillboxes can be appended*/
+}
+
+#clear-pillbox-holder {
+    position: absolute;
+    z-index:10;
+    top: 50px;
+    flex-grow: 1;
 }
 
 .main-view {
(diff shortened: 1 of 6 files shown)


ostermanj:
@NealHumphrey The demo's looking great. I'm back from my trip and hope to get back into the code soon. When are you leaving?

NealHumphrey:
@ostermanj my flight is out of Dulles at 10am tomorrow, so will be wrapping up in the next hour or two. I'll probably do a bit on the planes tomorrow too, but by then I'll likely stay away from the #344 checklist and just do some Buildings page cleanup since that's still in a pretty raw state. 

I'm working on a few of the smaller items in #344 now, if you start anything before I sign off just ping me on Slack, otherwise I'll leave a note on #344 with where things stand - mostly by checking off the checkboxes of everything I've done. 

For the most part I think we won't need more API changes for now so should be all front end, which can be deployed to the live site with a simple merge to the master branch.

ostermanj:
Cool. So if we commit improvements to dev before Tuesday we should then merge to master for the user testing?

NealHumphrey:
Yep. I'll make sure you and @ptgott have permission to push to master - that way you can either a) do a pull request and have the other one approve it, or b) do the merge to master locally and push to master (i.e. to deploy on your own without the other person needing to approve). a) is preferred option for QA/transparency but b) in case it's needed.  I'm assuming your master is following the codefordc one without extraneous edits.

PR #355, opened 2017-06-17
344-Small Tweaks

ptgott:
I'll be pushing to this pull request while addressing smaller issues in #344 today.

### DIFF
--- a/docs/tool/css/styles.css
+++ b/docs/tool/css/styles.css
@@ -319,14 +319,16 @@ a.active {
     padding: 3px;
 }
 
-.mapboxgl-ctrl-top-left{
-    top: 80px;
+.mapboxgl-ctrl-top-right{
+    top: 120px;
     position: absolute;
     z-index: 10000 !important;
-    transition: left .5s;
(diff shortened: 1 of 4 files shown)


NealHumphrey:
@ptgott Thanks for working on this! 

1) The back button fix looks good. 
2) The mapbox zoom control is close, but it has a couple small conflicts in its current position. 
- In collapsed filter mode, it's overlapping a little with the toggle-out button
- In open filter mode, if you add filters the newly added pillboxes are stuck underneath the mapbox control 

What do you think the best solution to this is?  What if we move the zoom control to the bottom left (instead of top left) and shrink the width of the buildings legend on both sides a little bit to accommodate the control and still looked centered?

ptgott:
@NealHumphrey Thanks for looking over this! I forgot all about the new filter pillboxes when moving the zoom controls. They're now on the top right, out of the way of the tab toggle and reset zoom button.

NealHumphrey:
Thanks @ptgott  - merged!

PR #357, opened 2017-06-19
Puts overlay layers below the project layer so that circles are always on top.

ostermanj:
I wish all fixes were so easy. mapBox addLayer method can specify which layer the new added one should be before. The line now specifies "projects." Overlays now all underneath projects so long as the project layer is loaded before someone selects an overlay. That should be the case.


ostermanj:
re: #344

NealHumphrey:
Great, thanks John!

Issue #334, opened 2017-06-12
Improvements to chloropleth legend


NealHumphrey:
Just a few tweaks to enhance the usability and performance of the chloropleth legends:

- [x] When toggling the data sets, the legend is added after the semantic UI accordion triggers, which leads to a jumpy 'pop' into place. It'd be good to make it smoother. Easiest technique is probably to add a fixed-height div just under the `id=overlay-about-DATANAME` div and append the chloropleth legend to that instead of to the `overlay-about-DATANAME` directly. Then the accordion will already know about the fixed-height div when it starts loading. 
- [x] The map has opacity but the legend does not, leading to a  visual disconnect between legend color and map color.  (EDIT I fixed this while I was fixing a merge conflict; switched the MIN_COLOR and MAX_COLOR to use rgba). 
- [x] (optional) This might be tricky, but it would be great if we could add some logical rounding of where the breaks occur so they happen cleanly based on some regular interval. So instead of crime:
    ```
    0-1,723
    1,724 - 3,447
    3,448 - 5,171
    5,172 - 6,895
    6,896
    ```
    we could do 
    ```
    0 - 1,500
    1,501-3,000
    3,001 - 4,500
    4,501 -6,000
    6,000+
    ```

    It would also be good to allow more than one entry in the top category, especially when there are more categories like neighborhood currently. In above example, the 6,896 category applies only to one ward.

PR #366, opened 2017-06-24
334 improve chloropleth legend

ptgott:
Addresses #334

### DIFF
--- a/docs/tool/js/core/housing-insights.js
+++ b/docs/tool/js/core/housing-insights.js
@@ -296,6 +296,11 @@ var setSubs = controller.controlSubs.setSubs,
 
  //TODO should these be wrapped up into a 'helpers' namespace? Or maybe the new utils section?
 
+// HELPER roundTo
+Math.roundTo = function(start, tensExample){
+    var numberToRound = start/tensExample;
+    return Math.round(numberToRound) * tensExample;
+}
 
  // HELPER get parameter by name
  var getParameterByName = function(name, url) { // HT http://stackoverflow.com/a/901144/5701184
--- a/docs/tool/js/views/chloropleth-legend.js
+++ b/docs/tool/js/views/chloropleth-legend.js
@@ -8,7 +8,7 @@ var chloroplethLegend = {
 
     chloroplethLegend.tearDownPrevious();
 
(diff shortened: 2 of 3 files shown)


docs/tool/js/core/housing-insights.js
```diff
@@ -296,6 +296,11 @@ var setSubs = controller.controlSubs.setSubs,
 
  //TODO should these be wrapped up into a 'helpers' namespace? Or maybe the new utils section?
 
+// HELPER roundTo
```
NealHumphrey: For this and the comma-separated thing, I realized that D3 already provides this functionality - we should use the d3.format() function to do this work for us. https://github.com/d3/d3-format

docs/tool/js/views/chloropleth-legend.js
```diff
@@ -36,11 +36,9 @@ var chloroplethLegend = {
           return x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ","); //adds thousands separators, from https://stackoverflow.com/questions/2901102/how-to-print-a-number-with-commas-as-thousands-separators-in-javascript
```
NealHumphrey: Again, using d3.format() would be cleaner - not blocking this PR unless you want to change it now, but noted for future refactoring.

docs/tool/js/views/map-view.js
```diff
@@ -135,9 +135,23 @@
...
+                    default:
+                        return Math.round(val);
+                }
+            }
+
             this.stops = new Array(CHLOROPLETH_STOP_COUNT).fill(" ").map(function(el, i){
                 var stopIncrement = MAX_DOMAIN_VALUE/(CHLOROPLETH_STOP_COUNT - 1);
-                var domainValue = Math.round((MAX_DOMAIN_VALUE - (stopIncrement * i))*1000)/1000; //stupid javascript rounding. Close enough for this purpose. 
+                var domainValue = MAX_DOMAIN_VALUE - (stopIncrement * i);
+                var domainValue = roundedVersionOf(domainValue);
```
NealHumphrey: do we want the second `var` here? Does it even make a difference to re-var the variable?

ptgott:
Thanks for your review @NealHumphrey! I'll merge this now and leave an issue for later.

PR #368, opened 2017-06-26
WIP - Add bar chart

NealHumphrey:
This pull request adds a bar chart prototype to our 'charts' section, and demonstrates how to use the bar chart in an example /tool/test_chart.html page. It has not been integrated into our tool yet, but this can be used in many places, probably starting with our right-hand sidebar. 

@ptgott and @ostermanj I especially want to talk about a few code structure things that I explored with the creation of this reusable bar chart:
- For this chart, I mimicked John's prototypical inheritance model of charts; but, I made a new base chart that does not perform the data callback. Instead, this chart-prototype expects that the data has already been loaded and is provided as `chartOptions.data`. The reason for this is flexibility - if we decide we want something slightly different than what the API provides, or want the user to be able to edit the data in some way such as selecting the denominator in a "percent of" calculation, this lets us perform those data manipulations before passing the data to the chart. In my example, I took raw numbers and calculated a percent field. I vote for making this pattern the norm for future reusable charts. Even though I think the preference should still be to have the API provide what's needed as-is, it can often be slow to get new API endpoints added for use. 

- I added a `chartOptions.percentMode` attribute that switches how the bar chart behaves (adds gray bars underneath, formats the axis text). Is this type of approach the preferred behavior, or would we be better off splitting this into another level of inheritance as was done with the donut chart that inherits from pie chart?

- I also created a Constructor version of the chart (I actually started with this one) to test out Mike Bostock's suggestion in his "Towards reusable charts" blog post. He does the data binding beforehand and passes in a d3 selection - while the constructor was a useful way to make the chart initially, I found that it lead to some confusing patterns needed to perform proper updates of the data. It also doesn't allow inheritance of partial charts. I think we're better off sticking to the prototypical inheritance. But, one thing to think about adding to our prototypical objects is the custom getter-setter methods. This allows for method chaining, which is the way the rest of D3 works - instead of using the chartOptions object on initialization (see width, height, container, etc. in the constructor version).


ostermanj:
A couple thoughts: we probably do need the flexibility of passing in custom (non-dataCollection) data into the chart constructors, so decoupling the constructor from the dataRequest sounds right, or at least we should make it optional somehow. When referencing the API-provided data object, even if it's to manipulate into something custom, though, we should code that reference as if we don't know if the data's been fetched yet or not (even if, really, we do) by using getData() rather than referencing the dataCollection directly. That way, the code is temporally flexible. If we decide to put it on the first view, it's ready to go; or is another developer copy-pastes it, it will work no matter when it's executed. I ran into that problem the other week when I moved some code that another module depended on for fetching the data it needed but tried to reference directly.

Second: the method chaining is really great and clean, but it seems it's not so much about making a chart more reusable as it it about making it more easily updatable, which of course is good too.

NealHumphrey:
@ptgott and @ostermanj we didn't get a chance to discuss this more last night, but based on your comments so far:

a) Let's definitely go ahead with this pattern for the data callbacks. Good point on using getData(), John - I'll include a note in the comments about that in a demo usage example. I'll go ahead and do a refactor of the donut chart to inherit from this prototypical chart instead, and/or put an obvious 'deprecated' note on the former ChartProto instructing people to inherit new charts from the new static version. 

b) I'll go ahead and add the most obvious method chaining attributes (height, width, etc.) but have them co-exist with the params method temporarily. We can look at the two versions and discuss which pattern we want to use going forward. As John notes, it's probably equivalent on initial chart creation but chaining might be a simpler way to do updates.

NealHumphrey:
Let's also make sure to all watch out for the getData() pattern instead of the direct data object reference in our code reviews, so that we don't slip in any direct references accidentally.

ostermanj:
We could consider refactoring the data objects to be fetch as ES6 Promises,
which I recently began wrapping my head around. That way, you'd *have to*
access them as if you don't know if they've been resolved or not. And it
would native-ize a lot of the logic built into the getData() method, with
the drawback that everyone would have to get comfortable with Promises.

NealHumphrey:
How do we feel about browser coverage on this? **https://caniuse.com/#feat=promises**

ostermanj:
IE11 is likely a big problem, but I think it's entirely polyfill-able since
it's an API addition rather than a change in syntax.

NealHumphrey:
@ptgott and @ostermanj 
I went ahead and refactored my ChartProtoStatic to make it the default prototype to inherit from, and started making some improvements to it. Changes include:
- Renamed the static (no-callback) one to be `ChartProto` and called the old one `ChartProtoCallback`. I have not refactored Donut/Pie/Timeline charts yet to use the updated method, though - wait until we're sure of new approach
- I rearranged how the proto methods and chart-specific methods interact with each other to be (hopefully) more clear. 
    - The chart prototype has `setup`, `resize` and `update` functions. Setup is for initializing default values, resize for anything that needs to change when changing the SVG size (including the first time you run it) and update for changing the content of the chart (including the first time)
    - Every chart-specific extension should have a `_setup`, `_resize`, and `_update` method. This are called at the end of the public prototype methods. Using the `_` to indicate a "private" method is a Python convention that's less common in Javascript I think but hopefully still clear.
- I added method chaining. Because we're not using a closure like they do in the 'Towards Reusable Charts' article, this results in a little bit of a clunky problem of not being able to use the same variable name for the function and the value itself - you can't store the value on `this.width` if `this.width` is a function that gets/sets the width. I again used the underscore convention, i.e. `this._width`. But, anytime you actually want to use the value, it's better to use the getter itself, e.g. `this.width()` which will return the stored value of `_width`. This way if we want to use defaults in the getter it will work. 
    The only downside to this setup is that anything that you don't use a getter/setter for you should access normally, i.e. `this.svg` which I used to store the D3 selection of the SVG DOM object. But, the it's just the person writing the chart that needs to know which things are regular `this.varname` instead of `this.varname()` - the end user of the chart will only deal with things that have getter/setters written for them. 

My current code I have tried to clean up / document all this stuff in the code comments. I plan to also make a quick new chart template (basically my BarChart file with a bunch of stuff stripped out and a few additional instructions/comments) so that people can more easily get up to speed on this approach. I also plan to add some nice utilities like automatic-resizing when the browser size changes for any chart that inherits from the ChartProto. 

Thoughts/feedback welcome!

NealHumphrey:
I'm merging this, based on our discussion so far - though the final version I pushed yesterday+today could of course use a second set of eyes. Probably best to fix anything that comes up as we go with whoever makes the next inherited chart!

PR #371, opened 2017-06-28
314 add geocode

jkwening:
Added methods to Cleaner module so that geographic zones in project table are encoded with data instead of nulls. I utilized api calls to MAR and added a cluster_desc_map dictionary object to prec_cat model. Since I was doing MAR api calls I also populated status for buildings that are unique with respect to those available in the prescat - that is, those that don't have a matching nlihc_id.


NealHumphrey:
This has been run on the AWS database so changes should now be available on the front end

PR #376, opened 2017-06-29
DEMO (but also ready to merge): Projects color coded by whether they match filters

ostermanj:
Projects are grayed out when they do not match filters. Projects that do not match the previous filter applied are distinguished from others so that people can see the results of their last action.

### DIFF
--- a/docs/tool/css/styles.css
+++ b/docs/tool/css/styles.css
@@ -131,19 +131,7 @@ html, body {
     padding: 10px;
     position: absolute;
     bottom: 0;
-    left: 260px;
-    right: 360px;
-    min-width: 235px;
-}
-
-#legend-items {
-    display: flex;
-    justify-content: space-between;
-    flex-wrap: wrap;
-}
-
-#legend-items div {
-    width: 310px;
+    left: 260px;    
 }
 
 #legend-items div span {
@@ -152,6 +140,10 @@ html, body {
     height: 10px;
     margin-right: 5px;
     width: 10px;
+}
+
+#legend-items div span.square {
+    border-radius: 0;
 }
 
 .sidebar-inner-wrapper {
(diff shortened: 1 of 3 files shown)


NealHumphrey:
@ostermanj thanks for jumping on this, good stuff! Functionally it works great. I do think that leaving the non-matching projects on the map in a gray state does solve multiple usability issues - it helps with when users have a leftover filter that they forgot about and are surprised when they expect a building to be in the filter they're currently working with, or expect to see a building that they don't see (happened with Danilo and Claire). It also helps draw better conclusions about the difference between selected buildings and the whole district. 

I agree with @ptgott that the styling needs some work to better understand the "doesn't match previous" filters. We should definitely discuss this on Thursday, and/or do some targeted user testing to see if people understand it or not. Some thoughts:
- Maybe use outlines in some way? So inner color is orange or gray indicating matches / doesn't match. Then a border color indicates 'different from previous setting'
- If we're indicating non-matching buildings that are new, we should also indicate matching buildings that have been added. Again maybe outlines could do this. It could be a neutral color like black for both of them, or maybe just a dark version of either color (gray or orange respectively) to indicate changes. 
- Showing the previous filter might be confusing if we don't also indicate what the previously selected filter was -you can remember it for a moment probably, but especially if you're playing with a slider for example you might not remember what exactly the previous value was. We should discuss how this might be included during the discussion of how to fix the usability problems surrounding filter activation that we saw with pretty much everyone (clearing filters, which ones are activated, etc.)
- Seeing this, another thing that I think of is that it also makes the graphic less helpful if you don't care about the previous state. For instance if we wanted someone to share (via our updating URLs) or export, they might only want to see filtered vs. not filtered. Maybe we should make this toggleable? Which then raises the question of where to put such a toggle and whether people will understand it. 

I may try out a few style edits to this branch (switching to using outlines) and see how that looks and then we can bring this to the planning discussion on Thursday (sorry you can't make it John!)

NealHumphrey:
I made some color edits to this PR and pushed it. Since I don't have access to @ostermanj 's repo I made it into a new PR. Closing this one in favor of that one (which has this code as the previous commit)

PR #377, opened 2017-06-30
314 add geocode

jkwening:
@NealHumphrey I think I've gotten them all now. The only one I couldn't figure out was pbca - which we don't seem to have any data to populate. Let me know if I missed something. For street view url some values were coming up as None in mar api call so I looked up the api and built a custom encoding based on lat/lon information available.

### DIFF
--- a/python/housinginsights/ingestion/Cleaners.py
+++ b/python/housinginsights/ingestion/Cleaners.py
@@ -295,12 +295,13 @@ class CleanerBase(object, metaclass=ABCMeta):
         full_address = row['Proj_addre']
         image_url = row['Proj_image_url']
         street_view_url = row['Proj_streetview_url']
+        psa = row['Psa2012']
 
         # only do mar api lookup if we have a null geocode value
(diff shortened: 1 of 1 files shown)


NealHumphrey:
Great, thanks @jkwening ! 

PBCA is "Performance-Based Contract Administrator Program property", which is a HUD thing. Since the Preservation Catalog aggregates the available HUD data, I would assume that if it is a property that is not in the HUD data and is therefore only getting funding from the DC Gov that we can assume that PBCA is 'No'. 

@ptatian - can you confirm that we can assume that pbca is no for any DHCD properties that are not already in the Preservation Catalog?

NealHumphrey:
Actually in making an example for the street view problem I noted a bigger issue with potentially incorrect lat/lons coming out of the MAR. That's a bit higher priority - but not related to this PR. Just heads up @jkwening

ptatian:
Yes, only HUD project-based rental assistance projects are potentially PBCA. This does not include tax credits, either, so anything not already flagged as PBCA is not PBCA.

PR #378, opened 2017-07-01
374 dchousing csv

jkwening:
Refactored dchousing api module to use batch download of csv file rather than via api call.


jkwening:
fixed - in reviewing my changes I forgot to push the additional code change.

Issue #380, opened 2017-07-02
End-to-end testing to anticipate the shift to V2


ptgott:
As we plan new features for V2 and think about overhauling the existing interface based on the user tests, I thought it would be good to include some end-to-end testing. I had originally looked into doing JavaScript unit tests with Jasmine, but ultimately sided with end-to-end testing because it seems like a nice way to specify a set of requirements for the tool while also (even if crudely) indicating that we're meeting them. In theory, end-to-end testing would be easy to make sure that by adding a new feature or conducting a major refactor, we haven’t wiped out any other features without realizing it. 

The ideal would be both unit tests and end-to-end tests, as the latter indicate broadly that a feature has failed in some way while the former can tell us where exactly in the code something has failed. For now, we could cover a basic laundry list of features with end-to-end tests.

PR #381, opened 2017-07-02
380 - Set up Capybara for end-to-end testing on a basic Sinatra server

ptgott:
This deals with issue #380.

To run the tests, first make sure Ruby and Bundler are installed and run 'bundle install'. Then go into docs/test and run 'rake spec'. 

For testing, I've used RSpec and Capybara. The decision to use these technologies is based on the fact that we're already using Ruby for Jekyll. Capybara starts up Selenium WebDriver, which interacts with a browser as specified in the tests. There's a Sinatra server that returns the contents of the 'tool' directory just for the tests.

One thing that'll be tricky with this approach is testing the Mapbox maps themselves. This is because Capybara's built-in matchers look for the state of the DOM, while the Mapbox maps are JavaScript canvases. I see two ways of dealing with this. First, we could delegate Mapbox-related testing to unit tests (e.g. the Jasmine Ruby gem), leaving anything related to the canvas out of the end-to-end tests. Second, we could use Capybara's 'evaluate_script' method to access the Selenium browser object and its associated JavaScript state. This second approach seems promising, but I haven't tried it yet.


ostermanj:
Thanks for setting this up. I like how it sounds but I won't be able to
review until I'm back from traveling next Sunday.

NealHumphrey:
@ptgott this is great stuff! I'm glad you got this rolling for us - we'll definitely want it for V2. Your choices sound good to me - but I've never actually set up a testing system in Javascript yet, so I'll rely on your judgement. 

I'm focusing on pulling together all the demos / strawmen for the Thursday discussion so probably won't dig in to review until next week as it'll take me a bit of learning to get up to speed on these tools. Is that ok? Do you think it needs reviewing/merging before then, or is that blocking you from taking it further at all? Let me know and I can rearrange if needed. 

Neal

ptgott:
@NealHumphrey Thanks for taking a look! So far I've only added a few tests, so I can keep working on adding more while pushing to this PR. Since the test directory is separate from the project source, I don't imagine there will be conflicts if we put off the review/merge.

PR #386, opened 2017-07-09
Generate more accurate street view urls

jkwening:
Initial code had incorrect street view url. Added new code using Google
Map API documented end point and street view validation end point to
produce closest street view url within 50 ft/meters if possible.

Note - there are still some locations with missing street view urls.
These are due to the given lat/lon for their location which point to
buildings that actually don't have valid google street view data. Some
of this is due to buildings not being on an actual street but behind
other street facing buildings.

@NealHumphrey how do you want to handle these? This may be related to what looks like incorrect lat/lon values. They don't seem to be incorrect, the error looks to be due to precision - at times the lat/lon is the center of the building hence why it differs to google maps at times; even their lat/lon data is inconsistent across their different api end points.

Apart from a few handfuls which legit don't have street view pano ids - the rest we I may have to create manual reference for them to load into the data base so this may not be perfectly automatic but could be reported as part of the logging component of the cron process.


NealHumphrey:
@jkwening I made a new issue for the missing street views. I think we talked about this briefly, but in any case let's put that off until later since it only affects a few buildings and we can fix it later down the road.

Issue #405, opened 2017-07-12
Incorrect longitude in one project


ostermanj:
The project with nihlc_id of 79575674-9b1b-48e2-aaff-ccfa84fd5523 (2700 16TH STREET NW) has  a positive longitude value. It should be the same abs value but negative (-77.036879).

Issue #397, opened 2017-07-11
Mapbox zoom when filtering


NealHumphrey:
Anytime a filter is selected, zoom the map view to the bounding box of any items currently included, zooming in or out as necessary. This will make changes to filters occuring outside of the current zoom more apparent, as well as selections of specific zones more aligned with user behavior (most expected it to zoom them to that area). 

To make this work, will likely need to:
- find the bounding box by finding the min and max lon of all the matching projects (note minLon will be the most negative longitude)
```
lowerLeft = minLat , minLon
lowerRight = minLat, maxLon
upperRight = maxLat, maxLon
upperLeft =  maxLat, minLon
```
- find the center of the bounding box - set this as center of the map
- find zoom level that matches bounding box (need to investigate how to do this in mapbox).

NealHumphrey:
Completed in #408! Thanks @ostermanj

PR #408, opened 2017-07-12
Map zooms in response to filtered projects

ostermanj:
re issue #397 

Subscribes to filteredProjectsAvailable state change; calculates bounding box of matching projects and zooms to fit the bounding back. mapBox, FTW, includes a `fitBounds()` method that handles this.

Includes workaround for issue #405 that can be removed when that issue is resolved.

### DIFF
--- a/docs/tool/js/views/map-view.js
+++ b/docs/tool/js/views/map-view.js
@@ -36,7 +36,8 @@
                     ['previewBuilding', mapView.showPopup],
                     ['filteredData', mapView.filterMap],
                     ['hoverBuildingList', mapView.highlightBuilding],
-                    ['filterViewLoaded', mapView.initialSidebarState]
+                    ['filterViewLoaded', mapView.initialSidebarState],
+                    ['filteredProjectsAvailable',mapView.zoomToFilteredProjects]
                 ]);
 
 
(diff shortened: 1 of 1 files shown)


docs/tool/js/views/map-view.js
```diff
@@ -994,5 +995,27 @@
                     'filter': ['==', 'nlihc_id', data]
                 });
             }
+        },
+        zoomToFilteredProjects: function(msg, data){
+            if (getState().filteredProjectsAvailable.length < 2 ) {
```
NealHumphrey: @ostermanj what do you think about removing this condition? If we run this including the first time, this helps resolve the issue related to Ward 8 being hidden from view - the current initial load state cuts off a tiny bit. This is less of an issue now that we have a smaller legend, but depending what we end up with for the layers legend we can use the padding of this function to make sure that none of the project dots are ever hidden  by the legend.

ostermanj: I wanted to avoid the "Reset zoom" button coming up on load, but, yeah, if we set the original zoom and center to match the fitBounds of all projects, we could remove it.

NealHumphrey:
With your loosened criteria, this ended up not showing the reset zoom button as is. 

Longer term, since we might end up adding or removing projects as the data gets updated, we might want to have the original zoom and center be calculated wherever they are originally set

PR #410, opened 2017-07-12
WIP:  Api v2

jasonrhaas:
## What's new
This is the initial API V2 overhaul.  It uses SQL Alchemy's `automap` feature to reflect the current database tables, and then uses Flask RESTLESS to serve up the API.  I have not ported over the `application.py` endpoints yet but that is the next step.

What this means is that we can now take advantage of all the nice built in features of Flask RESTLESS since we have an ORM model in place.  See the docs for more details:  https://flask-restless.readthedocs.io/en/stable/customizing.html

## Testing

In order to make this happen I needed to add a new column to the database called `id` and make it the primary key.  The SQLA requires that you have a primary key for the tables, and this is needed for Flask RESTLESS.  What this means is the database needs to be rebuilt to use this new API.  

To test:
- `docker-compose down` to clear out the docker images and database
- `docker-compose up -d`
- `cd python/scripts`
- `python load_data.py docker` to rebuild the database
- `cd ../api`
- `python ../api/declarative_app.py` to start the new API

At this point you can navigate to `localhost:5001/api/<tablename>` to use the built in Flask RESTLESS endpoints.

## Examples

`http://localhost:5001/api/crime`

Will yield paginated results of everything in that table.

`http://localhost:5001/api/crime/00002652-8648-456f-ae30-cb3d43d9273b`

Will should just that crime record.

`http://localhost:5001/api/crime?page=2`

Will show page 2 of the results.

You can also adjust results per page (default is 10)

`http://localhost:5001/api/crime?results_per_page=100`


python/api/declarative_app.py
```diff
@@ -0,0 +1,67 @@
...
+    manager.create_api(model, methods=['GET'])
+
+
+@app.route('/')
+def hello():
+    return("The Housing Insights API Rules!")
+
+
+if __name__ == '__main__':
+    app.run(host='0.0.0.0', port=5001)
```
NealHumphrey: Note, for this to work in AWS the variable `app` needs to be named `application`

python/api/requirements.txt
```diff
@@ -1,13 +1,33 @@
-# Note this should be synced with the environment.yml so that we can also run ingestion code on the server. 
-
```
NealHumphrey: Update: I got the declarative_app.py running by using my local Python install and pip-installing just the Flask-Restless and Flask-SqlAlchemy. But, I couldn't get the Docker Python Sandbox to create the environment with these new dependencies added. @jasonrhaas you added the sandbox in #245 - do you know how to get Docker to recreate the environment using an updated requirements.txt located in the root /python folder? We'll want to get that resolved before releasing this to the other developers or we'll have lots of system config questions pouring in.

docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
         restart: "on-failure:10"
 
     postgres:
-        image: codefordc2/housing-insights-postgres
+        image: mdillon/postgis
```
NealHumphrey: What was your motivation for changing this postgres image? I was able to get it running on my machine when I switched back to the existing postgres, though I only tested the crime endpoint. the `mdillon/postgis` image uses Postgres 9.6, while we're still on 9.5 - switching versions is ok if it's needed b/c AWS [now supports 9.6](https://aws.amazon.com/about-aws/whats-new/2016/11/amazon-rds-now-supports-postgresql-9-6-1/) but it will cause problems with anyone that is using their local Docker to run the database - they'll have to delete the associated volume b/c the existing data clashes.  If we can keep the existing image that would be easiest.

python/api/models.py
```diff
@@ -0,0 +1,353 @@
+# coding: utf-8
```
jasonrhaas: Yea I autogenerated it using a python tool.  Originally I was going to try to convert everything to the ORM model, but then I realized that in order to do that, the full ingest process, table creation, etc would have to be done in SQLA.

I can just delete the file to avoid confusion.

python/api/requirements.txt
jasonrhaas: Its probably due to the top level `requirements.txt` file not being updated.  Updating that now.

python/api/declarative_app.py
jasonrhaas: yup, that is true.  I'll re-name it.

docker-compose.yml
jasonrhaas: This was something owen changed way back when and I just never changed it back.  I think at one point we were considering taking advantage of some `postgis` capability, but maybe we never went that route.

I'll change it back to the original one since there doesn't seem to be any reason to use the `postgis` image right now.

jasonrhaas:
@NealHumphrey Everything should be good now.  Try doing `docker-compose down && docker-compose build && docker-compose up -d` to rebuilt the docker containers.

At that point run `docker-compose exec sandbox bash -c 'cd /repo/python/scripts && source activate /opt/conda/envs/housing-insights/ && python load_data.py docker'` and it will re-build the database.  

Then you can go into the sandbox and run the app from there if you want to (though it may not be shared outside the container).

jasonrhaas:
Also for the merge conflict, it should be resolved since I deleted the **python/api/requirements.txt** and am only using the **python/requirements.txt** and **python/environment.yml** files now.

NealHumphrey:
Allright, this is merged in. I'm going to take a stab at first pass restructuring the application.py and then we can look into bringing the declarative_app.py into that (maybe w/ blueprints?).

Issue #395, opened 2017-07-11
Continuous filter improvements


NealHumphrey:
Improve the UI and add functionality to the `continuousFilterControl` . This should also be done to the date filter being created in PR #385 

- Remove tooltips
- Add text entry boxes below the slider for min and max
- While the user drags the slider ("update" method of slider), continuously update the value that is typed into the relevant entry box. Example (but with only one slider....):
```
//create slider code not shown
var displayField = document.getElementById('field');
exampleSlider.noUiSlider.on('update', function( values, handle ){
	directionField.innerHTML = values[handle];
});
```
- When the user clicks the 'submit' button, move the slider and apply the filter
- When the user changes the value of the entry box and the form field loses focus, move the slider and apply the filter (adding this as a second trigger means the user doesn't need to click submit button. But, they still need the submit button since they need a way for the form field to lose focus when only editing one value)

- Add an 'include nulls' toggle button. For user clarity, change the text associated with the slider when it is toggled, to indicate current status i.e. "Unknown values included" or "Unknown values removed". 
- Update the filter functionality to accept a third value in the state array for continuous filters: `value=[min,max,includeNulls]` and properly filter the projects. 

![image](https://user-images.githubusercontent.com/5642566/28079554-dbc741b8-6636-11e7-885e-506d2dd0468e.png)

NealHumphrey:
Closing this issue as completed in #395 (except lingering bug noted in separate ticket)

Issue #387, opened 2017-07-10
URL state encoding


NealHumphrey:
After selecting several filter parameters, I want to be able to bookmark or share my findings by sending a URL to someone. When changing views in the single page app (such as changing from map view to look at a specific building) I also want my back button to behave as expected. Both of these things necessitate encoding the state of the app in the URL. 

Core functionality:
- [ ] Every time the user modifies a 'saveable state element', modify the browser url with `history.replaceState`
    - [ ] Change a filter
    - [ ] Switch sidebar type
    - [ ] Expand a sidebar
    - [ ] Select a specific building for preview mode
- [ ] Every time the user modifies a 'browsable state element', modify the browser url AND back-button history with `history.pushState`
    - [ ] Switch to building view
    - [ ] Switch to map view

# Proposed Implementation
## Status quo:
Currently our the 'state' variable (accessed via `getState()`) holds all the necessary information about the app state to recreate a specific view. However, it also holds a) other unnecessary stuff, such as 'dataLoaded' events, b) large data sets that are based on other state elements, like the filteredData object, and c) the previous state of each component in a 2 element long array. As such, we'll need to separate the state elements that are necessary for the URL from those that aren't. 

Once we have the relevant values, we need to be able to translate back and forth between the URL and the state itself. 
- First, we'll need to turn the current urlState into a url. To start with, the [jQuery.param()](http://api.jquery.com/jQuery.param/) function looks like the easiest way to do this. However, it will be quite possible that users could occasionally end up with very long URLs - potentially exceeding the [suggested limit of 2000 characters](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers). IE is the real limiting factor here, for which we already have spotty support but need to not ignore completely. We can manage this a little bit through careful attention to the length of both variables and values, such as mapping 'Subsidy Program' names to a key index. Longer term we could store states in the database if complex urls are common. 
- Then we'll need to turn a URL into a restored state. We'll want to be careful that if lots of similar events are triggered (such as each of the selected filters) they don't re-run the relevant code multiple times which can cause performance issues - this is the same problem that is currently occurring with the 'clear filters' button that triggers the recalculation of the filtered buildings multiple times. We also want to make sure the application of the setState() for these url params occurs at the right time - for instance, we can't apply filters if dataLoaded.filterData hasn't triggered yet. 

## Proposed architecture:
[...]

PR #411, opened 2017-07-14
395 Control date filters via text input

ptgott:
Addresses #395.

1. I've included a constructor for an input element that changes the state of a filter, something we should be able to use between date filters and continuous filters.
2. I've opted for input boxes over a calendar-style date picker because it is faster and more direct, assuming that the user doesn't need to know the day of the week a date falls on.
3. The filter input constructor allows the user to bind the input to a callback, which fires when 'tab' or 'enter' are pressed, or when the input element loses focus.
4. Setting the state of the date filter values is bound to the input element, not the slider.
5. Changing the slider changes the input element's year and triggers its callback.
6. Changing the input element changes the state of the slider without triggering its callback.

This subsumes #385


ptgott:
Looking over #395 again, I've noticed that the last few commits neglect some implementation details. I hadn't given the issue a second look before making the commits, and can make the changes this weekend if you'd prefer.
 1. They add the includeNulls key to dataChoices, rather than the state object. I've done this out of convenience, because the date pickers also change properties of dataChoices, taking it as an object that reflects the state of the data between the initial load and its use by the UI. There's probably a better object to use for this. I can move includeNulls to the state object, as intended, but I'm not sure about where to put the re-assigned min and max of the dataChoices objects of type 'date'.
2. There's no submit button. It'd be easy to add one.
3. The text by the switch is slightly different and the switch handle is two-dimensional.

NealHumphrey:
@ptgott Sorry it took me a couple days to get to this. 

Looks like good progress and getting close. I agree, though, that we should go ahead and make the changes you note. 

You should not use the dataChoices object for any user interactions - we should put everything that can be changed by the user onto the State object. a) this makes it cleaner to separate configuration from user changes and b) we'll definitely need this for #387 - URL State Encoding. 

The 'min' and 'max' values in dataChoices were originally used to determine the outer bounds of the sliders. As it appears you've done for the date slider this should actually be set by checking the data itself to find the min and max values. Note that the continuous number sliders have not yet implemented this functionality and still use dataChoices.max and dataChoices.min for their slider configuration. 

The best place to put this is on the state object with they keyword `filterValues.my_source_name` where my_source_name is the value stored in the 'source' field of dataChoices - the unique identifier of the data being filtered on. In the ticket I suggested passing a triplet of data instead of a pair with `setState('filterValues.portfolio', [minDateSelected, maxDateSelected,includeNullsBoolean])`. 

Currently the filter pillboxes are not working in this branch - clearing the filter doesn't clear the pillbox, and clicking the x on the pillbox doesn't reset the filter. That fix may be part of the setState issue. 

Final couple issues you note should be easy -submit button, null value text, and pointer cursor in css. 

Thanks Paul!

ptgott:
@NealHumphrey 
The latest commit has addressed the issues re: storing the nullValues state and touching dataChoices. I'll get to the cosmetic stuff/submit button later in the week.

NealHumphrey:
Great, thanks @ptgott! I'll wait to get the cosmetic stuff before doing a final review/merge, let me know when that's ready.

ptgott:
@NealHumphrey The newly formatted date/continuous filter controls should be ready now!

NealHumphrey:
@ptgott this is great! It's almost ready, there's just a couple things I'd still like to adjust before it gets merged in. 

- [ ] The biggest change - the choice to include/exclude nulls should actually be part of the filtering choices. Currently, if I select a filter such as Subsidy Start Date with the slider, and then I click the toggle so that unknown values are excluded, I end up with some filtered buildings. Then, I click the 'clear' pillbox and that filter disappears - from the pillboxes there is no evidence I have an active filter. But, I still have gray buildings that do not match my filter, because my 'clear' pillbox does not clear the choice I made. This is confusing, and its not clear how to get back to the reset state. 
    - [ ] Clearing the filter should reset the null value setting to its default state (included)
    - [ ] Toggling the null value filter to exclude nulls should trigger the creation of a pillbox to indicate which data choice is filtered via nulls
The cleanest way to do this might be to use a third value in the filterValues state ([min,max,null]) - as you noted there are many places in the code that assume that an empty filterValues means no filters are selected, but this is actually the behavior we want. If you want to do it a different way and keep your separate null values state that's fine too. 

- [ ] The null values toggle currently isn't very usable. I probably shouldn't have specified a slider toggle because despite their ubiquity they have pretty bad UX performance - they're confusing. Should we just change this to a checkbox?
- [ ] Change the css for text-input margin-left to be 20px (date picker submit button currently wraps when there is a scroll bar shown, while the 20px margin fixes it). 
- [ ] in dataChoices.js, remove all the 'min' and 'max' settings - since these are now calculated in the code based on the data, we don't need them in the configuration data so should clean it up by getting them out.

NealHumphrey:
Also - do merge in the current dev branch to resolve your merge conflicts so it's ready to get into dev

NealHumphrey:
@ptgott it looks like you've checked off all the checklist - is there anything left that you were planning to do on this?

All the changes look good. I did find a little bit of UI bug, though, related to the syncing of the UI and the actual filter state. I made a separate ticket for it. https://github.com/codefordc/housing-insights/issues/442

Let me know if there's anything else you expect still needs changing. I think that I will probably merge this in as-is, and let you/someone pick up the issue #442 in a separate PR so that we can use this new ui in a few integration related tasks.

ptgott:
@NealHumphrey Thanks for taking a look at this!

Issue #402, opened 2017-07-11
Scheduled CRON jobs on server for data collection


NealHumphrey:
We want our data ingestion and processing to occur automatically, on a regular schedule
- Check for new data weekly, or on a data set-specific schedule, or on demand when requested by admin
- Download and ingest all new data (will require refactoring to make it work in server environment)
- Handle all data ingestion problems via logging and skipping (many things currently can crash the code if something goes wrong)
- Alert admins via email of the data collection status - any missing/skipped data sets, any data ingestion problems.

speedyturkey:
I'd like to start working on this ticket if I can. I have a couple of questions which I'll post to Slack.

speedyturkey:
Putting this back for the time being, seems like it isn't as ripe as some other tickets.

speedyturkey:
I'd like to start taking a look at this. After discussion with @jkwening , it sounds like we'd benefit from some centralized logging functionality, which I have some ideas on.

Ultimately, it seems that the scheduled job (however it is scheduled) will need to call the code in get_api_data.py.

NealHumphrey:
@speedyturkey awesome, yes! Go for it. When we finalize the #400 issue we can get the scheduling code into the right place, but until then setting it up to be able to call get_api_data and do proper logging would be a big start. We especailly will want any warning-level logging to be sent on to the admin so we'll want to separate out regular from warning (or higher) logging in some way

NealHumphrey:
Closing in favor of more recent #470 that has more details and overlaps w/ this

Issue #399, opened 2017-07-11
Design project view


NealHumphrey:
We need lots more information in the project view (formerly called building view) Before implementing this, we need to design what this view will include. 

Note, some basic additions are being added in #398

ableo57:
@NealHumphrey please find an updated draft here: https://app.moqups.com/neal@nhumphrey.com/zeZOXaXpNx/edit/page/ad3aea595

ableo57:
Notes from my conversation with Neal:

![image](https://user-images.githubusercontent.com/24680097/28395320-25237d00-6cc1-11e7-8862-2d7a19f1f72f.png)

NealHumphrey:
@ableo57 Just want to check in on this - how's design coming? Let me know if you need anything, hoping to turn this into an action plan at tonight's session!

ableo57:
@NealHumphrey 

There are a few activities going on:

1. @txomashernandez is working to send out a survey to refine which metrics users would find helpful in the project view. Essentially, we're presenting an exhaustive list of variables and having respondents rate their importance on a likert scale

2. Dashboard graphics -- I've mocked up a few different options for both contextual and time-evolved graphics. I'm working to think through others as well.

3. Putting it all together/grouping: Thomas mentioned Jaime would be the best to put together the final organization of the mockup. Hoping to do that tonight, but if Jaime isn't able to come, I'll work at it and take care of it by the end of tomorrow.

Let me know if we've got developers chomping at the bit.

ableo57:
Neal, please feel free to comment this up: https://app.moqups.com/neal@nhumphrey.com/zeZOXaXpNx/edit/page/a5dfad0e0

PR #412, opened 2017-07-14
WIP: Refactor api

NealHumphrey:
Using Jason's style: 

# What's New:
- I updated the `.elasticbeanstalk` config to use the Code for DC AWS account
- `application.py` has been moved up one level to the /python folder. I tested deploying this to the Code for DC account; the copy of the API that is currently in `master` branch has been deployed to this new API. After our front end code is migrated to use this new API url I will shut down the one that is currently on my personal account. New API URL: http://housinginsights.us-east-1.elasticbeanstalk.com/

- I started splitting up the folder structure so that we can separate logically separate items into different files, to open up development for #402 . 

# Testing
Crank up your Docker instance, hop into the sandbox with `docker-compose exec sandbox bash`, `cd /repo/python` and run `python application.py`. 

Interesting urls to visit and compare to the code:
`localhost:5000/api/demo`: demo of a normal Blueprint
`localhost:5000/api/constructed_demo`: Returning a blueprint so that the blueprint can access any passed variables. This was mostly necessary for getting the database engine into the other views. 
`localhost:5000/api/count/crime/all/12/ward` This is the only pre-existing endpoint that I actually moved into a new file. The rest still need to be refactored. 

# Considerations
The constructor methodology was the best solution I could find to handling the fact that all the database endpoints need the database engine, and we don't want to reinitialize that in every single blueprint. But is this the best method?

What is the best logical separation of API code? With the addition of our ZoneFacts table we'll have fewer complicated endpoints. Does it make sense to group them based on which view in the housinginsights tool they will be used in? 

# Next Steps
- @jasonrhaas , @jkwening and @speedyturkey  in particular, let me know any feedback on the direction this is going. 
- I propose starting out by:
    - Dumping all the existing endpoints into one file inside /api called something like 'V1_endpoints` or `custom_endpoints`. 
    - Turning the declarative_app into a blueprint and calling it something like 'raw_endpoints'. This will end up replacing the 'raw' endpoint we have currently, though that will necessitate a bit of front end code restructuring due to the renaming of 'items' to 'objects' since objects is the key used by sqlalchemy.  
    - Putting scheduled jobs into its own module inside /api that gets imported into application.py however makes sense - I'm not sure if blueprints are relevant to scheduled flask jobs. 

@speedyturkey if you want to start on #402 by starting your branch off this one we're probably close enough to final format that a separate file for scheduled flask jobs that gets hooked up to application.py is probably safe to go ahead and work on and can be plugged into this approach.


jasonrhaas:
@NealHumphrey I'm getting an error when I try to run `python application.py`

```
  File "/Users/jhaas/miniconda3/envs/housing-insights/lib/python3.5/site-packages/psycopg2/__init__.py", line 130, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: FATAL:  password authentication failed for user "USERNAME_HERE"
FATAL:  password authentication failed for user "USERNAME_HERE"
```

jasonrhaas:
@NealHumphrey Nevermind I figured it out by looking at the code.  `python application.py docker` is what I needed.

jasonrhaas:
Got this response on this endpoint: `localhost:5000/api/count/crime/all/12/ward`

```
{
  "data_id": "crime_all", 
  "grouping": "ward", 
  "items": null, 
  "notes": "Query failed: (psycopg2.ProgrammingError) relation \"crime\" does not exist\nLINE 4:                 FROM crime\n                             ^\n [SQL: \"\\n                SELECT COALESCE(ward,'Unknown') --'Unknown'\\n                ,count(*) AS records\\n                FROM crime\\n                where report_date between (now()::TIMESTAMP - INTERVAL '12 months') AND now()::TIMESTAMP\\n                 \\n                GROUP BY ward\\n                ORDER BY ward\\n                \"]"
}
```

jasonrhaas:
I'm also wondering if we should separate out the custom endpoints and "raw" Flask Restless endpoints also.  I know the current API uses the standard postgres `psycopg2` statements and some python code to create the endpoints.  But if we put them in the same "application" we could potentially take advantage of the SQLA syntax for creating new custom endpoints if we desire it.

NealHumphrey:
@jasonrhaas - re: the crime endpoint, I just tested and did not get that error. Does your local copy of the database have the crime table in it? Have you rebuilt lately? 

Re: separating endpoints. I was thinking that it would be useful to logically separate stuff for the api into sections as our API codebase grows. In implementing it I did find that it was a bit more complicated than I expected due to access to configuration stuff in the different files, like the database engine and, as you point out, the SQLA syntax you've set up. It's solvable using Blueprint constructors as I demo'ed, and if people just copy the approach they don't even really need to think much about how it works, but it does make it more complicated. 

I'm open to one single application.py file, or at least just one for endpoints at least (with separate files for maybe utils and cron jobs). If so we'll just need to logically organize the function definitions with appropriate section breaks. It would make it easier to share config stuff like the SQLA table definitions. @jasonrhaas , @jkwening , @speedyturkey and anyone else w/ Flask experience - please weigh in, what are your opinons? We should pick an approach ASAP though so we don't have to reshuffle.

jasonrhaas:
@NealHumphrey I guess it depends if we are planning on running two separate API applications?  Is the idea to have them both running on elastic beanstalk as separate endpoints?  Or was that the reason for the `blueprint`?  I have never used `blueprint` before but it looks like its just a template to create endpoints?  Does that mean that we can have separate files running on the same actual application?

NealHumphrey:
Just one API and one server. My thought was breaking it up into separate files just for logical convenience / separation of concerns. 

Blueprints are one option for how to break an application into multiple files, and they seemed like the easier one to implement and deploy to AWS. They have some other functionality too for reusing code for similar endpoints, but we probably won't use that .

The current version of the code is one API application and it pulls additional endpoints in using blueprints. SEe [line 115 of application.py](https://github.com/codefordc/housing-insights/pull/412/files#diff-3e93ba97adaabb0cf9cf97ae6fff579dR115)

jasonrhaas:
How do we want to break this up?  Let's stick with your initial suggestions @NealHumphrey I think the Blueprint is a good idea.

NealHumphrey:
@jasonrhaas Let's start out by getting this into `dev` branch if it all looks good. I tested locally, but it would be good to get a formal code review on it. Can you check it out and run it locally and make sure you can get all the endpoints to work?

Steps:
1) @jasonrhaas or @jkwening If one of you does code review and approves of what's in this PR and verifies it works, go ahead and merge it into dev. 
2) @jasonrhaas refactor the declarative app into a blueprint and register it on the app
3) either @jasonrhaas or @nealhumphrey (whoever gets to it first - Jason, would be great if you have time, should be quick) move all the existing endpoints out of application.py and into a blueprint. I think we start with just putting them into a single 'old_endpoints' blueprint and then as we decide which new endpoints that will be added as the design for the Project Page (#399) gets finished we'll have a better idea of how to structure them. 
4) @nealhumphrey will do an assessment of existing and future endpoints to sketch out a rough plan of what all we'll need to add (this will probably be my top priority assuming Jason can take on 3). 
4) @speedyturkey you can plan for this structure for #402 - not sure if scheduling jobs works as a blueprint, if it needs to be in the app itself, or if a normal module structure works for it.

jasonrhaas:
@NealHumphrey I tested this again after reloading the database and it looks good.  I'll take a look at the code real quick.

jasonrhaas:
@NealHumphrey It looks like I'm not authorized to merge the pull request.

NealHumphrey:
@jasonrhaas - not sure why it wouldn't let you merge this - it's supposed to be mergeable by anyone with write permission after someone approves the review. In any case, I've explicitly added you to the list of people that can merge into dev now so should be all clear in the future

Issue #393, opened 2017-07-11
Make activated filters and filter clearing more usable


NealHumphrey:
In user testing, all the users got confused about which filters were activated and how to clear existing filters without clicking through each one individually. We need to design a different UI for showing activated filters and clearing all filters. 

Ideas for what we could do (some or all of the below):
- [x] Maybe move the filter pillboxes to appear in between the "2) Data Choices" label and the first data item?
![filter-pillbox-location](https://user-images.githubusercontent.com/5642566/28078252-3b306c74-6632-11e7-8603-b0bbfd702eef.png). Clicking the pillbox itself could activate that data set in the menu below and scroll the user to it, while clicking the 'x' could clear the filter. 
- [x] Either instead or in addition, have a 'view active filters' link (in that same area) that expands the div (pushing the data source list downward) and renders the text-readable version of the active filters that is described in #389 . 
- [x] Move the 'clear all' pillbox to be next to the individual clear filter pillboxes. 
- [x] Add an "x" on the right side of each activated filter title that serves as a second clear filter pillbox? This way users don't need to open/close if they want to clear.

NealHumphrey:
Completed in #414 - closing!

PR #414, opened 2017-07-15
393 - Place 'clear filter' labels at the top of the Data Choices menu

ptgott:
This PR also animates the 'clear filter' labels to fly from their associated UI components to their place among the other 'clear filter' labels. The aim here is to make it unmistakable that the orange labels emerge as a result of choosing filters, and that they move to the top of the Data Choices menu. 

Addresses #393

### DIFF
--- a/docs/tool/js/views/filter-view.js
+++ b/docs/tool/js/views/filter-view.js
@@ -29,7 +29,7 @@ var filterView = {
                         .data(activeFilterIds)
                         .classed("not-most-recent",true);
 
-        var allPills = oldPills.enter().append("div")
+        var newPills = oldPills.enter().append("div")
             .attr("class","ui label transition hidden")
             .classed("clear-single",true)
             // Animate a label that 'flies' from the filter component
@@ -66,14 +66,14 @@ var filterView = {
             });
 
         //Add the 'clear' x mark and its callback
-        allPills.each(function(d) {
+        newPills.each(function(d) {
             d3.select(this).append("i")
                 .classed("delete icon",true)
                 .on("click", function(d) {
                     d.clear();
                 })
-        })
-        .merge(oldPills);
+        });
+        var allPills = newPills.merge(oldPills);
 
         oldPills.exit().remove();
 
--- a/docs/tool/js/views/filter-view.js
+++ b/docs/tool/js/views/filter-view.js
@@ -3,33 +3,30 @@
 var filterView = {
 
     addClearPillboxes: function(msg,data){
-    
-        for (var i=0; i < filterView.filterControls.length; i++){
-            var x = filterView.filterControls[i]['component']['source'] //display_name
-        }
 
         //Compare our activated filterValues (from the state module) to the list of all 
         //possible filterControls to make a list containing only the activated filter objects. 
         //filterValues = obj with keys of the 'source' data id and values of current setpoint
         //filterControls = list of objects that encapsulates the actual component including its clear() method
-        var activeFilterIds = []
-        var filterValues = filterUtil.getFilterValues()
-        for (var key in filterValues){
-            if (filterValues[key][0].length != 0){
-                var control = filterView.filterControls.find(function(obj){
-                    return obj['component']['source'] === key;
-                })
-                activeFilterIds.push(control)
-            };
-        }
-        
+        var filterValues = filterUtil.getFilterValues();
+
+        var keysWithActiveFilterValues = Object.keys(filterValues).filter(function(key){
+            return filterValues[key][0].length !== 0;
+        });
+
+        var activeFilterControls = filterView.filterControls.filter(function(filterControl){
+            return keysWithActiveFilterValues.indexOf(filterControl.component.source) !== -1;
+        });
+      
         //Use d3 to bind the list of control objects to our html pillboxes
-        var oldPills = d3.select('#clear-pillbox-holder')
+        var allPills = d3.select('#clear-pillbox-holder')
                         .selectAll('.clear-single')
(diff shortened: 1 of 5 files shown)


NealHumphrey:
@ptgott - that sounds pretty slick! I will be able to review tonight.

NealHumphrey:
@ptgott 

The animation definitely helps to clearly designate that the filters are a result of the user action. I'd be curious to hear what others think of this UI behavior - @ostermanj ? @ableo57 ? @thischrisoliver ? Here's a screenshot for easier viewing:
demo_animate_pillbox


I'm also doing a review with a couple bugs I found plus some usability tweak suggestions

docs/tool/js/views/filter-view.js
```diff
@@ -191,7 +220,7 @@ var filterView = {
         }
 
         this.isTouched = function(){
-            return slider.noUiSlider.get()[0] === c.min && slider.noUiSlider.get()[1] === c.max;
+            return slider.noUiSlider.get()[0] !== c.min && slider.noUiSlider.get()[1] !== c.max;
```
NealHumphrey: One bug I found: when you click the 'clear filters' pillbox, the function to recalculate the filteredData list is run like 25 times, which causes some serious responsiveness issues during clearing. I haven't fully investigated but I suspect that this change is the source - remember the 'isTouched' was originally added to make sure that the clear filter button didn't trigger too often. 

I don't see why this function was changed to `!==`, though technically now I think it returns the proper 'true' and 'false state per its name (isTouched = true if either the min or max is altered). If changing this, you should also change the code that uses the isTouched when deciding whether to recalculate the filters. 

Future note: the 'forcePublish' function and refactor to use the 'updateFilters' event subscription is a better long term solution to this issue, and is part of the URL encoding refactor described in #387. At that point `isTouched` might no longer be needed.

docs/tool/js/views/filter-view.js
```diff
@@ -32,8 +32,36 @@ var filterView = {
...
+                    flyLabel.style.left = destinationElement.getBoundingClientRect().left + 'px';
+                    flyLabel.style.top = destinationElement.getBoundingClientRect().top + 'px';
+                }, 1);
+
+                // Remove flyLabel after its transition has elapsed.
+                window.setTimeout(function(){
+                    flyLabel.parentElement.removeChild(flyLabel);
+                }, 1500);
+
+            });
```
NealHumphrey: Unfortunately because you didn't edit the line I can't put the comment in the right place, but I would like to put this on [row 28](https://github.com/codefordc/housing-insights/pull/414/files#diff-cb09ec43944a93232c7ef43b427c1e8fR28). 

Bug I found: 
When clicking around with trying multiple filters, sometimes the filter does not place the proper pillbox. I couldn't quite figure out the pattern to reproduce the issue, but try: 
1) Slide the 'total units' slider
2) slide the 'subsidized units' slider
3) clear the 'total units' pillbox with the 'x'

Expected behavoir: total units pillbox removed
Actual behavior: 'total units' pillbox remains and 'subsidized units' dissappears. 

It looks like the clear button is actually clearing the right filter (see the console log from the State module - it publishes an empty proj_units_tot).

In a related issue, if you keep clicking around adding/removing pillboxes you can also sometimes end up with a situation where the code adds multiple copies of an active pillbox that is not the one you are currently clicking (e..g ACS Median Rent is added 5 times). I couldn't come up with a good click pattern to recreate this, but after enough playing with adding/removing I was able to make it happen every time. I think the pattern was re-adding ones that were above the last one after removing some in the middle/end. 

I think that the problem here is with the data binding that is occuring between the pillboxes and the activeFilterIds list. Since there is no key function, we are currently binding on the order of the pillboxes and the order of the `activeFilterIds` list. 

My guess is that adding a key function that binds the data and pillboxes based on the 'source' field will solve this problem: 
```
.selectAll('.clear-single',function (d) {
  return d['component']['source'];
})
```

ptgott:
@NealHumphrey Thanks for your review! Here's a checklist summary of your comments. I'll start pushing commits as I work through them.

- [ ] Make the gray 'clear all' pillbox be completely clickable with no 'x'.

- [ ] Put an anchor div at the fly-to destination and have the 'label' be the actual pillbox. Or otherwise, avoid duplication during flying.

- [ ] Animate the height transition of the #clear-pillbox-holder to avoid the sense of the data selection UI jumping.

- [ ] Fix bug: When applying multiple filters, sometimes the filter doesn't place the correct pillbox, or removes the incorrect pillbox.

- [ ] Fix bug: When adding/removing pillboxes, sometimes the code adds multiple copies of an active pillbox (not the one currently clicking). This is likely due ot re-adding pillboxes above the last one after removing some in the middle/end. Check the data binding between pillboxes and activeFilterIds. 

- [ ] Change the 'merge' call around line 74 to the statement, 'var allPills = newPills.merge(oldPills);'

- [ ] Fix bug: When clicking 'clear filters', the function to recalculate filteredData 'is run like 25 times'. This might be down to the change in 'isTouched' and the code that calls it.

NealHumphrey:
@ptgott this looks great! It looks like all the issues are resolved. 

I did a couple small edits of animating the exit of the pillboxes with a fade-out to clean things up a bit. I tried to get the animate height transition to work as noted in the last bullet point but had some tricky issues - the method I mentioned works (setting max-height using d3, using a css transition on max-height to do the animation) but it was buggy to get the proper max-height to set using d3 because sometimes the pillboxes wrap and sometimes they are on new lines depending on their height. The fadeout helps with the jerkiness on the back end, so I say leave this alone for now and we can decide if we want to do it as a final cleanup step later down the line.

Issue #404, opened 2017-07-11
Add nlihc_id to topa table


NealHumphrey:
The TOPA notices table will sometimes correspond to records in our projects table - though there will be many additional TOPA notices for addresses that are not part of the projects we know about. To be able to join these two tables, we want to add the `nlihc_id` (which is the unique identifier for the project table) . Check the cleaning of the 'dchousing' data set for an example of how to do this.

NealHumphrey:
- Project table has a field mar_id which should match to the address_id from the topa table. 
- Any non-matching records can have null nlihc_id

speedyturkey:
Grabbing this one. Spoke to Neal and John and it will hopefully be an easy one (fingers crossed).

speedyturkey:
In speaking with @jkwening about this, it looks like the Cleaners do not currently have access to the database connection. My current plan is to have load_data pass the db engine to the base cleaner so that I can do the lookup on ID. That way, anyone who needs to do similar lookups at this stage in the future will be able to. Please let me know if this conflicts with your understanding of things.

speedyturkey:
An alternative implementation would be to perform this update all at once as a post-processing script. If we add nlihc_id to topa table we can do:

```sql
update t
set
  nlihc_id = p.nlihc_id
from topa as t
inner join project as p
  on t.address_id = cast(p.mar_id as numeric)
```

NealHumphrey:
@speedyturkey I'd probably vote for option 1, adding database access to Cleaners.py - that way we're consistent with the one place that we modify data.

speedyturkey:
Thanks for that guidance Neal. I made a WIP PR #415 with some further comments. In short, I have added some necessary functionality but could use some direction on how to complete the issue.

NealHumphrey:
Completed

PR #415, opened 2017-07-16
Add nlihc_id to topa table #404

speedyturkey:
I made some modifications which allow for the topa cleaner to verify existence of a record in the project table with a given address id (MAR id). It was a bit trickier than expected.

Given this functionality, I am not sure about next steps. My notes say that we want to add a new field called nlihc_id to the topa table and populate it if it exists, but wouldn't it just be a duplicate of the existing address_id? Perhaps a boolean field would be better so as not to duplicate data? I'm afraid I don't understand how the field will be used.

This PR is currently stable - so I am happy to keep working with further guidance, or to hand off to someone to polish off (I do think the remaining work should be the "easy" part).

Thanks as always.

### DIFF
--- a/python/housinginsights/ingestion/Cleaners.py
+++ b/python/housinginsights/ingestion/Cleaners.py
@@ -508,7 +508,7 @@ class dchousing_cleaner(CleanerBase):
 
         return row
 
-class topa_cleaner(CleanerBase):
+class TopaCleaner(CleanerBase):
     def __init__(self, meta, manifest_row, cleaned_csv='', removed_csv='', engine=None):
         #Call the parent method and pass all the arguments as-is
         super().__init__(meta, manifest_row, cleaned_csv, removed_csv, engine)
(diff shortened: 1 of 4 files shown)


NealHumphrey:
@speedyturkey The nlihc_id and address_id don't contain the same type of id - address_id should really be named to mar_id for consistency, and it's maintained by the DC government to indicate an address. nlihc_id is the unique identifier of a project. Technically some projects can contain multiple addresses and so have additional address_id (mar_ids) associated with them - though in the project table we just use the 'primary' one. 

It's useful to add the nlihc_id to the topa table because nlihc_id is the one we use to univerally identify projects in our database - while technically we could join the TOPA table to the project table directly using mar_id, enabling the (slightly duplicative) nlihc_id makes for a consistent user interface in how the data is stored across all the tables. 

I'd say yes just do a quick add of that field to the table in the cleaner. Note, to add the column you'll need to first add it to meta.json (which tells the code to create a blank spot in the database table for that field), and mark it as 'required_in_source=false' meaning it doesn't expect to find it in the csv file. Then add a line on the cleaner that writes to that column/field of the dictionary. Let me know if you have any issues with that.

speedyturkey:
Thanks a lot @NealHumphrey. This PR is ready now, all is working on my end.

python/housinginsights/ingestion/Cleaners.py
```diff
@@ -507,13 +508,32 @@ def clean(self, row, row_num=None):
 
         return row
 
-class topa_cleaner(CleanerBase):
+class TopaCleaner(CleanerBase):
+    def __init__(self, meta, manifest_row, cleaned_csv='', removed_csv='', engine=None):
+        #Call the parent method and pass all the arguments as-is
```
speedyturkey: Thanks @NealHumphrey.

I think the reason I did this was that if I created TopaCleaned.__init__ in order to pass the engine to the cleaner, it would skip CleanerBase.__init__. So the cleanest solution (at least, that I could think of) was to modify CleanerBase. A side benefit is that all classes inheriting from CleanerBase can now access the engine.

Issue #180, opened 2017-04-06
Weighting of census tract data


NealHumphrey:
Weighting files for translating from census tract to the other zone types have been added to the /data/geographic_data/tract_weights folder. 

We are waiting on #152 and for a longer list of fields from the census data before we need to implement this.  

Notes from Peter Tatian, who provided the weights files:
-------------------------------------------------------------
A few things to note. For apportioning counts of people, housing units, etc., you want to use the popwt field. For example,
 
 
**Geo2000**|**Ward2012**|**Pop**|**Pieces\_tr00**|**Pop\_tr00**|**Pieces\_wd12**|**Pop\_wd12**|**Popwt**|**Popwt\_prop**
:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:
11001000100|Ward 2|4674|1|4674|25|63455|1|0.073658
11001000201|Ward 2|4467|1|4467|25|63455|1|0.070396
11001000202|Ward 2|3850|1|3850|25|63455|1|0.060673
11001000300|Ward 2|2028|2|5706|25|63455|0.355415|0.03196
11001000300|Ward 3|3678|2|5706|21|75335|0.644585|0.048822

This shows the share of the tract (Geo2000) that is in the ward. So, 100% of tract 11001000100 is in ward 2, but only 35.5% of tract 11001000300 is in ward 2. The remaining 64.5% of the tract is in ward 3. In this case you would be doing a weighted sum of the tract counts to get the ward level.
 
(The Popwt_prop field is used for taking a weighted average of a proportion, mean, or median. Generally, I don’t recommend doing this, but sometimes you may not have a choice if you can’t create what you need from counts.)
 
There are versions of the weight files for both 2000 and 2010 census tracts.
 
We use a version of neighborhood clusters in NeighborhoodInfo DC that is slightly different from the ones published on opendata.dc.gov. These clusters were based on 2000 census tracts and so are a cleaner aggregating data. This was the recommended approach of the state data center at the time.

NealHumphrey:
@ajhalani I don't know why the data dissappeared from that folder (or maybe I wrote the ticket planning to sync it there but didn't...). In any case, I've re-uploaded the weighting files to S3. 

[...]

PR #416, opened 2017-07-17
WIP: 388 zone facts table

jkwening:
## What's new

1. Added a function that runs after all data loading is completed that drops the existing records from the zone_facts table and rebuilds the table.

1. Includes function that generates the values used in zone column ensuring that they match the relevant standard column name and zone_specific match the value for that column that is used in other tables.

1. Populate the zone_facts table with calculated fields from the "census_with_weighting" api endpoint: 'poverty_rate', 'fraction_black', 'income_per_capita', 'labor_participation', 'fraction_foreign', 'fraction_single_mothers'.

## Testing
I need to complete unit test code so it is more robust.

To test:

### DIFF
--- a/python/housinginsights/ingestion/LoadData.py
+++ b/python/housinginsights/ingestion/LoadData.py
@@ -314,6 +314,9 @@ class LoadData(object):
             # add to db
             metadata.create_all(tables=[zone_facts])
 
+            # populate table with calculated fields values
+            self._add_census_with_weighting_fields_to_zone_facts_table()
+
         return metadata.tables
 
     def _add_census_with_weighting_fields_to_zone_facts_table(self):
--- a/python/housinginsights/ingestion/LoadData.py
+++ b/python/housinginsights/ingestion/LoadData.py
@@ -295,31 +295,46 @@ class LoadData(object):
         return processed_data_ids
 
     def _create_zone_facts_table(self):
+        """
+        Creates the zone_facts table which is used as a master table for API
+        endpoints. The purpose is to avoid recalculating fields adhoc and
+        performing client-side reconfiguration of data into display fields.
+        This is all now done in the backend whenever new data is loaded.
+
+        :return: a immutable dict that contains the tables structure of the
+        table and sqlAlchemy metadata object used.
+        """
         with self.engine.connect() as conn:
             # drop zone_facts table if already in db
             if 'zone_facts' in self.engine.table_names():
                 conn.execute('DROP TABLE zone_facts;')
 
-            # create empty zone_facts table
-            metadata = MetaData(bind=self.engine)
-            zone_facts = Table('zone_facts', metadata,
-                               Column('zone', String, primary_key=True),
-                               Column('poverty_rate', Numeric),
-                               Column('fraction_black', Numeric),
-                               Column('income_per_capita', Numeric),
-                               Column('labor_participation', Numeric),
-                               Column('fraction_foreign', Numeric),
-                               Column('fraction_single_mothers', Numeric))
+        # create empty zone_facts table
+        metadata = MetaData(bind=self.engine)
+        zone_facts = Table('zone_facts', metadata,
+                           Column('zone', String, primary_key=True),
+                           Column('poverty_rate', Numeric),
+                           Column('fraction_black', Numeric),
+                           Column('income_per_capita', Numeric),
+                           Column('labor_participation', Numeric),
+                           Column('fraction_foreign', Numeric),
+                           Column('fraction_single_mothers', Numeric))
 
-            # add to db
-            metadata.create_all(tables=[zone_facts])
+        # add to db
+        metadata.create_all(tables=[zone_facts])
 
-            # populate table with calculated fields values
-            self._add_census_with_weighting_fields_to_zone_facts_table()
+        # populate table with calculated fields values
+        self._add_census_with_weighting_fields_to_zone_facts_table()
 
         return metadata.tables
 
     def _add_census_with_weighting_fields_to_zone_facts_table(self):
+        """
+        Populates the zone_facts table with the calculated fields data.
+
+        :return: list of SQLAlchemy query result objects from inserting the
+        calculated fields data into the zone_facts table.
+        """
         census_with_weighting_fields = [
             'poverty_rate', 'fraction_black', 'income_per_capita',
             'labor_participation', 'fraction_foreign',
@@ -398,7 +413,10 @@ class LoadData(object):
 
         Takes data that is formatted for output the API, i.e. a dictionary
         with key "items", which contains a list of dictionaries each with 'grouping'
-        and 'count'
+        and 'count'.
+
+        Returns items as dictionary with group and division result as
+        key/value pairs instead of list of dictionaries.
         """
         items = {}
         if numerator_data['items'] is None:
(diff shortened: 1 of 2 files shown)


NealHumphrey:
Thanks @jkwening ! I'll be able to review this tonight or tomorrow morning, and ideally merge it in before Tuesday's work session (I'll be virtual again this week).

jkwening:
@NealHumphrey it only took minor code changes to also populate the acs rent quartile data so I went ahead included it in this PR. The process for testing is unchanged.

Quick question: for my edification - why are we using 'population_weight_counts' instead of 'population_weight_proportion' for census to ward/neighborhood_cluster weight factor? I would have expected the latter. Thanks.

NealHumphrey:
@jkwening it actually depends on the data source which weighting factor we should use. [Here is the description](https://github.com/codefordc/housing-insights/issues/180) of which to use where. We'll probably need to add some sort of configuration variable for each data source telling it which weighting factor to use (suggestion where to put that welcome).  In short, any thing that is a total sum for the zone (i.e. total income, total people working, etc.) should use population_weight_counts and anything that is a median/average/etc. should use population_weight_proportion. On the UI we should also flag anything that is weighted using a proportion since it's not quite as statistically accurate so should be used only when the source data size is not useful.

python/housinginsights/ingestion/LoadData.py
```diff
@@ -286,8 +289,267 @@ def rebuild(self):
...
+        """
+        # drop zone_facts table if already in db
+        if 'zone_facts' in self.engine.table_names():
+            with self.engine.connect() as conn:
+                conn.execute('DROP TABLE zone_facts;')
+
+        # create empty zone_facts table
+        metadata = MetaData(bind=self.engine)
+        zone_facts = Table('zone_facts', metadata,
+                           Column('zone', String, primary_key=True),
```
NealHumphrey: Should add a column here for 'zone_type' which has 'ward', 'neighborhood_cluster', or 'census_tract' as the field value. This is for convience when getting the data out of the table later - instead of a sql statement with a `where in (big,long,list,of,zones)` we can just use `where zone_type='ward'`.

python/housinginsights/ingestion/LoadData.py
```diff
@@ -286,8 +289,267 @@ def rebuild(self):
...
+        return query_results
+
+    def _get_zone_specifics_for_zone_type(self, zone_type):
+        """
+        Returns a list of zone_specific values for a given zone_type.
+        """
+        with self.engine.connect() as conn:
+
+            if zone_type == 'ward':
+                table = 'project'
```
NealHumphrey: Would it make more sense for this table to be `census_tract_to_ward`?

python/housinginsights/ingestion/LoadData.py
```diff
@@ -286,8 +289,267 @@ def rebuild(self):
...
+
+        # populate columns accordingly for each zone_specific type
+        for zone_type in zone_types:
+            field_values = dict()
+
+            # get field value for each zone_specific type
+            for field in census_fields:
+                result = self._census_with_weighting(data_id=field,
+                                                     grouping=zone_type)
+                field_values[field] = result['items']
```
NealHumphrey: In this context, returning the dictionary that includes the 'items' key probably doesn't make sense since we'll be writing this straight to the database, not making a user interface. I'd suggest refactoring _census_with_weighting to just return the content of 'items' and then this field can fust be field_values[field] = result

python/housinginsights/ingestion/LoadData.py
```diff
@@ -286,8 +289,267 @@ def rebuild(self):
...
+        # If data_id isn't one of our custom ones, assume it is a column name
+        else:
+            api_results = self._get_weighted_census_results(grouping, data_id)
+
+            # transform items to dict with group/value instead of list of dict
+            temp_items = {}
+            for item in api_results['items']:
+                temp_items[item['group']] = item['value']
+
+            api_results['items'] = temp_items
```
NealHumphrey: As noted above, don't need to store this on the 'items' field anymore - that was just relevant for how the API needed to serve its results in json format

python/housinginsights/ingestion/LoadData.py
NealHumphrey: that said, might make sense to just let it lie for now since it's working and used in mutliple places....

python/housinginsights/ingestion/LoadData.py
```diff
@@ -286,8 +289,267 @@ def rebuild(self):
...
+                # item = dict({'group': n['group'],
+                #              'value': divided})
+                items[n['group']] = divided
+
+        return {'items': items, 'grouping': numerator_data['grouping'],
+                'data_id': numerator_data['data_id']}
+
+    def _census_with_weighting(self, data_id, grouping):
+        """
+        Zone facts table helper function to get data from our census table.\
```
NealHumphrey: Would be good to add to docstring: 'returns weighted data for either a) a column from the 'census' table or b) a calculated field defined by a keyword that has a custom calculation defined in this method.` or something liek that

python/housinginsights/ingestion/LoadData.py
```diff
@@ -286,8 +289,267 @@ def rebuild(self):
...
+                    count = 0
+                    for result in results:
+                        tract = result['census_tract']
+                        factor = result['population_weight_counts']
+                        matching_data = next((item for item in census_results if
+                                              item["census_tract"] == tract),
+                                             {field: None})
+                        if matching_data[field] is None:
+                            logging.warning(
+                                "Missing data for census tract when calculating weightings: {}".format(
```
NealHumphrey: Would be helpful to add which data field was missing in addition to which tract. When I ran this locally, I got 34 of these warnings but don't know whether to be worried about the mssing data or not.

python/housinginsights/ingestion/LoadData.py
```diff
@@ -286,8 +289,267 @@ def rebuild(self):
...
+
+    def _populate_zone_facts_table(self):
+        """
+        Populates the zone_facts table with the calculated fields data and
+        acs rent data fields from census.
+
+        :return: list of SQLAlchemy query result objects from inserting the
+        calculated fields data into the zone_facts table.
+        """
+        census_fields = [
```
NealHumphrey: For reference for future developers, should add a comment here describing the purpose of this list. Contains the list of field names to be included in the zone_facts table. Names should correspond to either a) a column name in the 'census' table or b) a calculated field as defined in the self._census_with_weighting function.

python/housinginsights/ingestion/LoadData.py
jkwening: yep, that was what I was thinking - I refactored only what I needed to make it work they way I wanted it to. As you will see, I changed how 'items' from list of dictionaries to a dictionary with group/value pair instead. I will simplify later if needed once everything is final output-wise.

python/housinginsights/ingestion/LoadData.py
jkwening: you're right - that would be better just in case in future a ward isn't represented in the project table.

python/housinginsights/ingestion/LoadData.py
jkwening: May be this is no longer necessary with this change but I was planning on including each 'zone_type' in 'zone' column as aggregate values. I was probably over thinking the notes in that bullet.

PR #422, opened 2017-07-19
WIP: url state encoding

ostermanj:
Very much work in progress.

re: #387 

Working so far only on URL encoding and decoding the states of the filters. Sidebar state and preview state not yet addressed. View switch not yet addressed.

**When a filter is selected:**
(router.js)
* `history.replaceState()` with the `stateObj` representing all active filters pushed to the window and a location hash pushed to the URL
* does not use `jQuery.param()` because the resulting string was too messy and verbose; needed custom pattern and abbreviations
* adds short_name property to dataChoices and uses that as the abbreviation 

**When a URL with location hash matching "#/HI/" is loaded:**
* a "Loading saved state" screen is placed over the view
* router.js parses the location hash and sets app state individually for each filter represented in the location hash
* is working for the functioning categorical and continuous filters

**Notes**
* router.js sets the window's `history.state` but in the end this is not ever read back in. the `history.state` would only be available on browser back or forward, not on a fresh load of a paramified URL, so I opted to rely only on the encoded URL for forwards, backs, and fresh loads. This might avoid some cross-browser issues, too.
* Reading in a paramified URL triggers the filters and everything works, **except** that the filter UI elements are not affected by the programmatic triggering of the filters. That's a to-do.
* Also, X'ing out a pillbox isn't working on filters read in from the URL. This is probably connected to point above. Maybe the X'ing out relies on DOM status (the positioning of the filters ) rather than state? Needs some investigation

**To do**
* somehow bind the filter UI elements to a programmatically triggered filtering
* add functionality for date filters, when available
* add sidebar status and preview status to the existing (replaceState) functionality
* build out similar handling for view switching, using pushState


NealHumphrey:
Sounds good so far @ostermanj. Let me know when you're ready for a review. 

On your TODO about binding the filter UI elements, I think you'll want to copy the `this.clear()` approach that is included in the creation of each of the filter control elements. You can add something like `this.set = function(values) { }` to each of the component types. Each of the filter UI elements requires its own custom way of adding values to it.

NealHumphrey:
@ostermanj Just wanted to check in - where does this stand? Let me know if it's ready enough for a review with current progress. If it's possible to merge in enough changes that we can get the back button to work when a user visits the project page that would be a good intermediate step - and would be worth including in the interim user tests happening at the end of this week. Let me know!

ostermanj:
I'll be able to work on this much of tomorrow and should have it all ready
before the work session. Hope so!

NealHumphrey:
Great, thanks!

ostermanj:
Continuous and date filters are being encoded and decoded successfully, in addition to categorical filters. I'll probably be able to commit the rest of the ticket (view changes and sidebar statuses) by the end of the night, but wanted to update the PR before the work session started.

ostermanj:
Getting there . . .  Latest commits handle back button functionality from the browser and in-app. A couple things still to come:

— encoding and decoding the sidebar states of the map view
— handling the include nulls boolean of the filters

NealHumphrey:
@ostermanj I've just done a small amount of test driving this branch, not a proper code review. 

- Back button for project view works as expected (yay!)
- Continuous and date filters work and are being appropriately reset in the view
- Categorical filters are only working for choices that do not have spaces/dashes in them. For example, for subsidy program, HOME and LIHTC work but "Project-based section 8" is not properly decoded when loading a url that contains it (it is placed in the url in the original view, though). 

Given the back button functionality and progress so far, I'm inclined to merge as-is so that the user testing can have the back button, and keep picking up the rest of this in a future PR (sidebar, categoricals with spaces, and nulls). What do you think?

ostermanj:
Great! I should have an updated commit before five, if you want to wait. I
was sooo close, but discovered some buggy behavior with the clear pillboxes
and clear all pillbox. I think I've almost got it, though.

ostermanj:
@NealHumphrey All set here. The underscore bug is fixed, plus some other messy issues I was dealing with. Still to come: sideBard state encoding/decoding, and removing url parameters when filters are manually set back to original conditions.

PR #423, opened 2017-07-20
WIP - Logging for get_api_data and sources

speedyturkey:
Implemented centralized logging for get_api_data and individual source ApiConns.

Currently work in progress / proof of concept for logging. I will try to reach out to @NealHumphrey or @jkwening regarding the best way to test this - want to make sure it is safe to execute all of these functions locally without any unwanted side effects.

Next step is to think about sending out the logs to an admin. My initial thought is to pick up the full log file at the end of get_api_data, attach it to an email, and then archive that log (safer than deleting).

### DIFF
--- a/python/housinginsights/sources/DCHousing.py
+++ b/python/housinginsights/sources/DCHousing.py
@@ -4,14 +4,15 @@ web site.
 """
 
 from pprint import pprint
-import logging
 
 from housinginsights.sources.base import BaseApiConn
 from housinginsights.sources.models.DCHousing import FIELDS,\
     DCHousingResult
 from housinginsights.sources.models.DCHousing import PROJECT_FIELDS_MAP,\
     SUBSIDY_FIELDS_MAP
+from housinginsights.tools.logger import HILogger
 
+logger = HILogger(name=__file__, logfile="sources.log", level=10)
 
 class DCHousingApiConn(BaseApiConn):
     """
@@ -48,14 +49,15 @@ class DCHousingApiConn(BaseApiConn):
 
         for u in unique_data_ids:
             if u not in self._available_unique_data_ids:
-                #TODO this will always be raised when passing a list to get_multiple_api_sources method for those not in this class. 
-                logging.info("  The unique_data_id '{}' is not supported by the DCHousingApiConn".format(u))
+                #TODO this will always be raised when passing a list to get_multiple_api_sources method for those not in this class.
+                logger.info("  The unique_data_id '%s' is not supported by the DCHousingApiConn", u)
 
             else:
                 result = self.get(DCHousingApiConn.DATA_URL)
                 if result.status_code != 200:
-                    err = "An error occurred during request: status {0}"
-                    raise Exception(err.format(result.status_code))
+                    err = "An error occurred during request: status {0}".format(result.status_code)
+                    logger.error(err)
+                    raise Exception(err)
 
                 content = result.text
 
@@ -67,4 +69,3 @@ class DCHousingApiConn(BaseApiConn):
                     self.create_project_subsidy_csv(
                         self._available_unique_data_ids[0], PROJECT_FIELDS_MAP,
                         SUBSIDY_FIELDS_MAP, db)
-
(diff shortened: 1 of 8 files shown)


NealHumphrey:
@speedyturkey sounds good! Are you ready for me or @jkwening to review this as-is? 

Can you describe what you mean by best way to test this? What kind of side effects are you thinking we should be watching out for?

On the sending of logs to admin - I agree on the described approach as the minimal version. But, it would be good to think about some sort of very basic level of flagging we could do to make sure the admin is paying attention when something occurs that needs their attention. It is likely just something as straightforward as  watching for warning or error level logging and then putting the fact that a warning or error level issue occured into the subject line and/or body. I just predict that the admin receiving this will see weeks and weeks of 'no issue' emails and start ignoring them. Can implement the email-log-file approach first but just flagging need for the next level in case there is anything you see while doing that version that you could implement as yougo.

speedyturkey:
No review needed at this time, although if you'd like to and offer feedback that would be fine by me.

In terms of side effects, I guess I mean API calls, downloading files, updating databases...I suppose it would just be my local DB so it's safe. I just get nervous running these kinds of scripts on my local machine when I don't fully understand all of the operations that are being done. From your response, it sounds like it is perfectly safe for me to go ahead and run during testing.

Thanks for the comments re: emails, makes sense. It should be possible to have the email notification present either a count of notifications by level, or to include something highlighting the most severe issues (for example to have the subject line clearly indicate whether a serious error took place).

NealHumphrey:
@speedyturkey running the get_api_data.py and the load_data.py should only affect your local machine and database copy, so there should be no problem running them to test your code. 

- get_api_data downloads the relevant files into the `/data/raw/apis/YYYYMMDD` folder on your computer with todays datestamp. If you are only doing it for testing purposes, you can delete this date stamped folder after the test code is run. If you are doing it 'for real' the aws cli sync command will make that data available to others; technically if two people run it on the same day and both run the sync command we'd have conflicts, but we're not intendeing to use sync long term (hence the cron jobs). 
- load_data.py takes all the files listed in `/python/scripts/manifest.csv` and adds them to the database that is specified. Database connection strings are stored in secrets.json. If you don't specify a database, the local database is the default. If you run it using the 'docker' keyword argument, the database in your local docker container is used. You have to explicitly choose the 'remote' keyword for it to use the live database. 

In short - run load_data.py with the 'docker' keyword (or code in a cron job that does the same thing as what is in load_data.py) and you'll be fine.

speedyturkey:
Thanks @NealHumphrey.

Unfortunately I am likely to miss tomorrow's session (bad work week).

I've pushed to this branch some basic functionality for emailing the logfiles - the new class, HIMailer, can be used for any kind of email, with or without attachments. To use it, I added a section to json.secrets:

    "email": {
      "username": "Email account name",
      "password": "Email account password",
      "host": "e.g. smtp.gmail.com",
      "admin": "Email address to cc on all emails (optional)",
      "debug_recipient": "I don't know if we want this, but allows you to override recipients when a debug variable is set to True. Good for testing."
    }

speedyturkey:
@NealHumphrey and @jkwening - I think this is ready for review. It's very much MVP in terms of the "attach and email log, let recipient know if there were errors" functionality. Some small tweaks would make it a little friendlier on ye olde ende user.

I am unfortunately very crunched at work over the next week, so I wanted to submit this so it can be merged, tweaked, etc as appropriate. If you have any feedback or if there is anything further that I can do, I will try my best.

Thanks!

NealHumphrey:
Great, thanks @speedyturkey ! I'll do a review and we'll get it in - we can build on the functionality more later.

NealHumphrey:
FYI this is now in the code! To run the get_api_data.py you'll need to add the email section to your secrets.json. @jkwening in particular, you'll want to edit your copy of secrets.json so that the 'admin' and 'debug_recepient' emails are both your email. In production, the admin email will be different of course.

```
"email": {
  "username": "housinginsights@gmail.com",
  "password": "Ask Neal for this",
  "host": "smtp.gmail.com",
  "admin": "youremail@yourdomain.com",
  "debug_recipient": "youremail@yourdomain.com"
}
```

Issue #391, opened 2017-07-11
Project preview mode


NealHumphrey:
On the 'Projects' sidebar, which currently provides a scrollable list of all matching projects, create a building preview section above the building list

- Project preview should occupy approx 1/3 of the height of the sidebar (~350px??). 
- Minimum Contents should be building name, address, ownership type. Add some snapshot charts showing things like unit count, bedroom mix, and reac score. This snapshot from an early mockup can provide some inspiration, but feel free to adapt as necessary. 
![image](https://user-images.githubusercontent.com/5642566/28076871-fd398986-662d-11e7-87b8-24e4c3daa518.png)
- Project name in preview should be a clickable link to the project (formerly building) view. 
- Remove 'owner' from the scrollable list, and add a 'view more' link in the scrollable list that links to the project view. 

- Change the building shown in the preview on a) click of the building in the scrollable list ( except the 'view more link) and b) click of the dot on the map
- Change the tooltip on the map to be shown on hover instead of click. Remove the link from the tooltip and instead just show the building name + address. Users can instead navigate to the project page via clicks on the sidebar.

NealHumphrey:
Completed enough, opening a new issue for enhancements per v1.5 user testing

PR #426, opened 2017-07-25
WIP 391 project preview

NealHumphrey:
I started on #391 adding a project preview mode. Still a WIP - mostly need to get the quick-view charts such as a REAC score in there. Progress to date:

- Added a 'preview' section at the top of the matching projects list
- Populated the preview section with basic building stats (i.e. assisted units, total units, address, etc.)
- I use meta.json to publish the display_name. This sets us up to later add a tooltip or something to show the display_text which is a short description of what the data means, as well as to link it to the data documentation page. 
- Refactored the donut chart to use the new chart-prototype. It is available to see in action on /tool/test_chart.html. This sets us up to be able to use a donut chart for the REAC score, but we first need to get 'most recent REAC score' available in the project data (this is a back end task, I can do this but heads up @jkwening). We will also want to swap out the existing donut charts on the 'summary' tab for these ones for consistency. 

Still todo:
- Make all relevant data available in the project table and/or in the geojson (REAC score, bedroom mix, etc.)
- Add charts such as REAC score chart
- Add link to project full page
- Change behavior of hover/click of the tooltip  on the map

Can either do a work-so-far review, or wait until it's finished.


NealHumphrey:
Here it is in action
demo_project_preview

NealHumphrey:
@ptgott I made the change we discussed re: height of the preview area vs. the 'matching projects' area - I think it works really well that way. If the screen is big enough, the preview takes 350 pixels and the matching list takes the rest; if the screen is too small, they split the space 50/50 and both get scroll bars. 

I also rearranged how the tooltips work on the map to use hovers, added the proper link to the project view, etc. It was a bit tricky to work through the mapboxgl.popup stuff but I think it works well now. @txomashernandez - one thing to pay attention to in user testing is if people are able to find how to view the project page of a specific project vs. just seeing its preview in this new arrangement - we might want to duplicate the link to the project page by putting it in the popup too. 

I'm merging as-is, but would be worth a quick spot check from either @ptgott or @ostermanj if you get a minute.

NealHumphrey:
Also, note next up I will be adding in the icon graphs - I had to get data into the APIs first. That was done in #435 that is a PR ready to review by someone back end.

PR #429, opened 2017-07-25
WIP: Refactors most of the current api endpoints

NealHumphrey:
into a few different blueprints, and registers them on the app.

Spot-checked the majority of endpoints so should be OK to use, need to deploy to server and put all the endpoints into the tool to make sure it all actually still works.


NealHumphrey:
@jasonrhaas Heads up, I went ahead and did this refactoring. Still need to fold in the declarative_app, which will allow us to delete the endpoint we currently have called `list_all()` (which I've left in application.py with a TODO to delete once done) . 

Next up I'm going to go ahead and shift some of our tables to use the new zonefacts table and add in stuff to the filter endpoint, which are the big things our front end needs, as I sort through what exactly we currently need on the front end. We'll end up dropping a few outdated endpoints that we can also swap out on the front end if we do it that way, so we don't need to worry about testing all the existing ones exactly until that's sorted out.

NealHumphrey:
*

jasonrhaas:
@NealHumphrey I just get this when running `python application docker`.  Is this expected?

```
INFO:root:['meta', 'manifest', 'building_permits', 'subsidy', 'project', 'reac_score', 'real_property', 'dc_tax', 'zone_facts', 'topa', 'census', 'census_tract_to_neighborhood_cluster', 'census_tract_to_ward', 'crime', 'wmata_dist', 'wmata_info', 'zone_housingunit_bedrm_count']
```

NealHumphrey:
@jasonrhaas I am pretty sure that comes from this line (line 90):
`logging.info(tables)`

Which is just echoing back the tables that it found when it connected to the database. Assuming so, yea that's expected (and was in there pre-refactor too).

jasonrhaas:
@NealHumphrey With my latest commit I added the Flask Restless Blueprints to `application.py`.  I added them into the application itself since it relies on an actual **application** instance and connection to the database to create the models.

Let me know what you think.

NealHumphrey:
@jasonrhaas This looks good. I agree putting the automap on application makes sense - and when we want to use the mapped models in any of our blueprints, we can pass them through via the blueprint constructor. 

As a side note, I find it safer to use the 'one branch per person' workflow - i.e., rather than pushing directly to the `400-move-to-blueprints` branch that I made initially, you make another branch off of that branch, and then do a PR either into the `400-` branch, or replace this pull request by making a PR into dev. It's slightly more branch management, but with a team of diverse Git skill level it's one less thing to think about - you can always push to a branch that you made without thinking about whether there are any upstream changes. Just something to keep in mind for future development. 

Merging in; I'm going to tweak a few things like pagination defaults after merging but will do that in another PR as I'm going to do the front-end changes to use the new endpoints concurrently so I can get them all working smoothly together.

Issue #430, opened 2017-07-27
Calculated fields on Project table


NealHumphrey:
Add a step to the code that updates the database to add calculated fields to the `project` table. Any project-specific data field that is contained in another table and that is used in the 'filter' endpoint should be added to the table. This means that to create the 'filter' endpoint, the CSV file for export, etc., we only need to join the `project` table to the `zone_facts` table and we will have all the data we need. This means some data is duplicated. 

Note that some fields from the Preservation Catalog are already calculated in this way - we should re-calculate these fields in case our additional data has altered their value (e.g. if a project in the PresCat receives a subsidy from DHCD that was not in the PresCat, the 'proj_units_assist_max' field might change). 

Typically, column names should take the form `<description>_<tablename>_<field_in_source_table>`. 
    - <description> corresponds to how the one-to-many relationship was resolved, i.e. "most_recent"
    - <tablename> is the exact table name from the database
    - <field_in_source_table> is the exact field name from the table that the value was pulled from. 

Every column that corresponds to a record in another table should also include the `id` field for that record in that table. For example, the `most_recent_reac_score_num` field would contain the score itself (e.g. 94), and a second column would contain the `reac_score.id` contents, named something like `most_recent_reac_score_id`. Note this is autogenerated using a uuid4 during the ingestion process, and so will not persist from database copy to database copy. However, assuming that the `update_project_table` script runs every time anything in the database is changed, this should always be the current copy of the uuid, that can be used for looking up the full record from the source table that corresponds to that data in the project table. 

*Note: pull request recommended any time a new field is completed*
MVP:
- [ ] Function that runs after 
- [ ] Most recent REAC score (part of #425)
    - `most_recent_react_score_num`
    - `reac_score_id`:

New Fields:
- [ ] Distance to nearest metro station (part of #424)
- [ ] Number of bus routes within 0.5 miles (part of #424)
- [ ] Most recent TOPA notice date (if one exists)
- [ ] `dc_tax.appraised_value_current_total`
- [ ] Appraised value per unit (appraised_value_current_total divided by proj_units_tot - note, should be at end of process so both numerator and denominator can be updated if needed)

Update existing calculated fields:
- Add 'id' records for all the current calculated fields listed below
- [ ] proj_units_assist_max (replace current value by looking in subsidy table)
- [ ] proj_units_assist_min (replace current value by looking in subsidy table)
[...]

Issue #425, opened 2017-07-24
Add 'Recent REAC Score' filter to map view


louvis:
Per issue #401, we want to add a filter for REAC scores to the map view that has a slider for the score number. I propose this filter pull only the most recent score to provide better insight on the building's current state.

The reac_score table includes 'nlihc_id' so it does not need to be added from the project table. Sample SQL code for the filter is provided below.
```
SELECT reac_score.nlihc_id
	, reac_score.reac_score_num
	, dates.last_score
FROM reac_score
INNER JOIN 
	(SELECT nlihc_id
		, MAX(reac_date) AS last_score
	FROM reac_score
	GROUP BY nlihc_id) AS dates
	ON reac_score.nlihc_id=dates.nlihc_id
	AND reac_score.reac_date=dates.last_score
```
The details required for the filter will need to be added to /docs/tool/js/views/dataChoices.js.  For both filters the `component_type` value should be set to `continuous`.

Since /docs/tool/js/views/filter-view.js already handles `continuous` components and data documentation exists for the reac_score table at /docs/data/reac_score.md, those steps are already completed.

The last step will be to add the field to the filter_data api endpoint, however since that code is in flux as of 18 Jul (related to PR #412 and PR #416)), check with @jkwening for an update before attempting this.

NealHumphrey:
The data is now available on the `raw_project` data set when using the new Code for DC API. PR today will switch front end API URLs to use this new endpoint and make this accessible. 

Still todo: 
- [ ] Add to DataChoices.js
- [ ] Review the pull request #430 (just for quality assurance - it is already merged)

Thanks @louvis for the SQL query - that was very helpful in making the code to edit the project table! Definitely valuable to have those sample sql queries for any additional tickets you make re: additional calculated fields

PR #435, opened 2017-07-27
430 calculated project fields reac

NealHumphrey:
Adds the functionality to have calculated fields in project table during updates of the database. Implements this for the 'most recent reac score'. Relevant to both #425 and #430. 

My only concern so far is maintaining the messy SQL statements required to extract the relevant data from the partner tables. I tried to implement the automap function so that I could try to use SQLAlchemy syntax, but I didn't see a way to then actually use that to do the stuff I wanted to. Straight SQL seemed easier to implement. 

At a minimum we could break each field update into a separate method - at the moment I have comments to designate sections of the _populate_calculated_project_fields code, but the could be methods that are listed out one by one just to break up teh code a bit. 

@jkwening especially your review would be appreciated!

I have deployed this to the Code for DC RDS instance - we'll be switching over to that one for V2 launch, though for the moment I'm using both the EC2 and RDS instances as sandboxes for testing our changes so we don't break the live site with stuff like this. Contact me for the connect string to add to secrets.json to check it out.

### DIFF
--- a/python/housinginsights/ingestion/LoadData.py
+++ b/python/housinginsights/ingestion/LoadData.py
@@ -247,6 +247,7 @@ class LoadData(object):
 
         # build Zone Facts table
         self._create_zone_facts_table()
+        self._populate_calculated_project_fields()
 
         return processed_data_ids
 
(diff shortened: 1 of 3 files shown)


NealHumphrey:
@jkwening heads up I'm going to merge this in as-is, but would still appreciate a second set of eyes w/ a review at some point. There's a bunch of loosely related back end api changes (zonefacts, refactored api, this one) that I'm going to try to consolidate and then swap out our front-end API urls to use all this new stuff for a map-view V1.5 launch early next week.

PR #436, opened 2017-07-28
WIP: 388 zone facts table

jkwening:
## What's new

Zone_facts table now includes data for the remaining census table columns, building permits, and crime. Also addresses some of the comments from #416.

## Testing:

## Notes
- crime and building permits currently uses 12 months counts because we don't have 2017 building permits data yet
- null values for census_tract zone_type for building permits data (see to do section for solution implementation)
- zone_type column doesn't yet include zone_type aggregate

## To do
- [x] add code that allows calculating census_tract values from ward and/or neighborhood_cluster data; will resolve null values for building permits data
- [ ] download and process building permits 2017 data into the database
- [ ] add code that includes zone_type aggregate in zones column
- [ ] connect zone_facts table to "filter_data" API end point and join project table accordingly
- [ ] complete unit test confirming validity of zone_facts calculated values
- [ ] complete get_residential_permits method that is called in summarize_observations method


python/housinginsights/ingestion/LoadData.py
```diff
@@ -312,6 +313,7 @@ def _create_zone_facts_table(self):
         # create empty zone_facts table
         metadata = MetaData(bind=self.engine)
         zone_facts = Table('zone_facts', metadata,
+                           Column('zone_type', String),
```
NealHumphrey: In thinking through how this gets eventually used on the front end - I think we want to instead create the ZoneFacts table using meta.json. At a minimum, even if it is created using the sqlAlchemy table syntax , we need to have the duplicate information in meta.json because that is where we have the 'display_name' and 'display_text' for each of the fields, which (will be) used for things like tooltips on the front end.

NealHumphrey: So, at a minimum: add a code comment here that says 'also update meta.json when adding/removing columns', and then add the appropriate meta.json fields. Or more cleanly, refactor this create table statement to use meta.json.

NealHumphrey:
@jkwening Thanks for the PR! I'm going to approve and merge this into `dev`, since it moves us forward on being able to implement related stuff (mostly via zone_type addition). But there's a few changes that should be made next, in addition of course to those you note in the todo section.

- I'm confused what you mean by "add code that allows calculating census_tract values from ward and/or neighborhood_cluster data". I saw the last commit is the one that adds code to deal with this - it looks like you are taking the ward number and dividing it by the number of census tracts in that ward. Is this related to the fact that that the building_permits table does not have a census_tract column? I think rather than trying to back-track that from the weighting factors, we should geocode the building_permits as a data cleaning step instead. Until we do that, I'd rather have null values for zones that we can't calculate for some reason. If for some reason we can't geocode the results, or in the future there are other fields that need to be similarly divided out in some way, we should be very explicit about this to the user so they know what kind of data interpolation has been done. Without this addition we can tell one story about how data gets reweighted to zones (we go from census tract and roll up to cluster/ward - never going to a smaller zone type). 
- This should be the same issue -  after running the code I see that many census tracts have the same value for 'building_permits' and 'construction_permits'. This shows it's clearly an interpolated value. 

- what does this mean? " add code that includes zone_type aggregate in zones column"

Otherwise looking good. Thanks!

python/housinginsights/ingestion/LoadData.py
jkwening: Will do but I'm a little confused. I thought one of the reasons we transitioned to Flask SQLAlchemy ORM was to replace meta.json by implementing models. I'm still playing catch up with all the work you and Jason have done but wanted to make sure I still understand what the end goal looks like. Thanks.

NealHumphrey: Hey @jkwening sorry for the confusion. After some back and forth, where we landed was: 
a) it seemed like too much trouble for not enough benefit to rewrite the ingestion functions to use the ORM - what's there now is working and there's not much different behavior that the ORM would provide
b) Jason found the Automap feature, which allows us to reflect the current database structure into ORM objects without directly mapping them
c) We would need to custom-build the display_name, display_text etc. stuff for the front end in some way anyways even if we refactored into the orm

So - we get some benefits in the API by using Automap and Flask-Restless can use it for some autogenerated endpoints. But, we'll keep the ingestion as is. AS I said fine to use the ORM for this table if you want but we should still put the table into meta.json for the display_text and display_name purposes anyways, so probably better to just do it wholesale. Either approach is fine though.

PR #439, opened 2017-07-29
WIP: 388 zone facts table

jkwening:
## What's new

PR changes requested by @nealhumphrey
- Don't back-calculate building_permits from Ward data. Zone_facts should report null values if a zone aggregation field (i.e. census_tract) is not available in the source table. So building_permits and construction_permits will have null values for census_tract until geocode building_permits to have census_tract data available during cleaning is complete. 
- Create zone_facts table using meta.json instead of sqlalchemy to match with other tables.

In addition, connected zone_facts table to "filter_data" API end point and joined project table accordingly

## Testing:
- run the sql query in filter_blueprint.py to confirm result is project table + zonetype_columnname for each zonetype

## Notes
- Modified HISql class in SQLWriter.py to allow directly creating table from meta.json without manifest row. It was a quick solution to the second request above. I think we should refactor the module to decouple methods that require manifest row object vs just having meta object or is there a better alternative I'm not thinking about?
- I couldn't confirm the api/filter endpoint returned the correct json output because I kept receiving error that zone_facts table doesn't exist. Also, my query involved multiple nested subqueries - I wonder if there's a cleaner sql query?

## To do
- [ ] download and process building permits 2017 data into the database
- [x]  connect zone_facts table to "filter_data" API end point and join project table accordingly
- [ ] complete unit test confirming validity of zone_facts calculated values
- [ ] complete get_residential_permits method that is called in summarize_observations method

### DIFF
--- a/python/api/filter_blueprint.py
+++ b/python/api/filter_blueprint.py
@@ -12,27 +12,68 @@ def construct_filter_blueprint(name, engine):
     @blueprint.route('/filter/', methods=['GET'])
     def filter_data():
         q = """
-            select p.nlihc_id
-              , p.proj_units_tot
-              , p.proj_units_assist_max
-              , cast(p.proj_units_assist_max / p.proj_units_tot as decimal(3,2)) as percent_affordable_housing
-              , p.hud_own_type
-              , p.ward
-              , p.anc
-              , p.census_tract
-              , p.neighborhood_cluster
-              , p.neighborhood_cluster_desc
-              , p.zip
-              , c.acs_median_rent
-              , s.portfolio
-              , s.agency
-              , to_char(s.poa_start, 'YYYY-MM-DD') as poa_start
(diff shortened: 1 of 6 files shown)


NealHumphrey:
@jkwening Changes look good. I think the quick solution is probably enough - it would be cleaner to refactor as you suggest, but I think there are other things I'd rather refactor / do in that time considering the stage of the project. 

On the API, I implemented a similar/related/conflicting approach to joining the zone_facts to the filter data set. I used a series of left joins rather than nested subqueries, and used some logic to generate the list of columns rather than hardcoding. It looks like your last commit has just the api changes and none of the other commits have api changes, so I'm going to make a new branch off of your second to last commit and merge that one in instead. I'll close this pull request then - it'll show as not merged, but it'll be everything except the API changes.

PR #440, opened 2017-07-31
Ready: 434 Slice zone facts

NealHumphrey:
- Replaces the current endpoint for census with a new zone_facts endpoint that can return any column from the zone_facts table

- alters the format of this endpoint slightly - next up will alter the front end to match

### DIFF
--- a/python/api/zone_facts_blueprint.py
+++ b/python/api/zone_facts_blueprint.py
@@ -7,123 +7,38 @@ import logging
 
 
 from api.utils import objects_divide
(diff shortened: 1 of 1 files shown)


NealHumphrey:
Completely removed the existing census endpoints, because they were calculating the same things on the fly that we now calculate more thoroughly in the creation of the zone_facts table. Endpoint to return data from zone facts is a simple query.

NealHumphrey:
Merging as-is so I can deploy and make necessary matching changes on front end - don't expect this PR to create any issues.

PR #443, opened 2017-08-03
WIP 392 - Fix location filters 

ptgott:
So far I've added a quick 'if' block to filterData() to filter by location. Next I'll work on clearing the existing location filter when the user changes zone type.

### DIFF
--- a/docs/tool/js/views/filter-view.js
+++ b/docs/tool/js/views/filter-view.js
@@ -114,6 +114,7 @@ var filterView = {
                 ['dataLoaded.filterData', filterView.formatFilterDates],
                 ['filterDatesFormatted', filterView.buildFilterComponents],
                 ['subNavExpanded.right', filterView.expandSidebar],
+                ['mapLayer', filterView.clearLocationBasedFilters],
                 ['mapLayer', filterView.updateLocationFilterControl],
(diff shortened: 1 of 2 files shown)


docs/tool/js/views/filter-view.js
```diff
@@ -978,7 +979,19 @@ var filterView = {
...
-
+    clearLocationBasedFilters: function(){
+        function isLocationBased(filterControl){
+            return (
+                filterControl.component.hasOwnProperty('zones') || 
+                filterControl.component.component_type === 'location'
+            );
+        }
+        for(var i = 0; i < filterView.filterControls.length; i++){
+            if(isLocationBased(filterView.filterControls[i]) && filterView.filterControls[i].isTouched()){
```
NealHumphrey: This is currently triggering every locationbased control, whether or not it was touched. With this we're having the performance problem of recalculating the filtered data 20+ times.

docs/tool/js/views/filter-view.js
```diff
@@ -978,7 +979,19 @@ var filterView = {
         }
         filterView.indicateActivatedFilters();
     },
-
+    clearLocationBasedFilters: function(){
+        function isLocationBased(filterControl){
+            return (
+                filterControl.component.hasOwnProperty('zones') ||
```
NealHumphrey: The part of the configuration in dataChoices.js that determines if it's a zone-based data set is
`"data_level": "zone",`

It would be better to use that property  instead of whether it has a list of zones.

Issue #392, opened 2017-07-11
Finish merged layer/filter


NealHumphrey:
The demo-v2-options branch has our new map view UI, which includes a merged layer/zone navigation for the data choices UI. The UI of this is mostly working but not all of the data is actually being applied to the filter view. 

- [x] Get the new combined 'location' selection to apply appropriate filters to the location
- [x] Changing location type will need to clear any selected 'location' filters (since previously selected choices are no longer relevant). Need to make this obvious to user -maybe use a tooltip to show that it has been cleared? Include a 'don't show again' button that removes the tooltip for the current session (this is much easier to implement than adding cookies which can be a long term add, if we keep this ui element). 
- [x] Merge filterView.components with mapView.initialOverlays for a unified list to be used for all views. After #388 is complete, we can move a lot of this out to meta.json - for now it can be either in code or in a static config file.
- [x] For now, flag any 'layer' items that do not work with filters yet - after #388 is complete we will have the necessary data available. 
- [x] Add the legend for layers when a layer-relevant data source is selected. Move this out of the data dropdown where it was in 'layers' and instead put it in the bottom right of the map.

NealHumphrey:
I'll be working on the first bullet points of this one today so we can close out this issue.

NealHumphrey:
Closing this - #443 resolves the first two bullets.

PR #444, opened 2017-08-03
WIP 392 zone aggregate filters

ptgott:
This PR addresses #392. It adds a constructor for filters based on aggregated data about a specific zone. When filters of this type are added, it checks the current map layer and, depending on whether that map layer is part of the filter component's 'zones' array, either adds the filter UI or displays a message that the filter is not available for the current zone. It also (ideally) calculates the filter's min and max values. This is just a start, and there are still a lot of runtime errors to deal with before we find out how much this works as intended. I'm throwing it out there so we can fix it as soon as possible. It's also a separate pull request from #443 because it's currently quite rough.

### DIFF
--- a/docs/tool/js/views/filter-view.js
+++ b/docs/tool/js/views/filter-view.js
@@ -397,8 +397,6 @@ var filterView = {
 
         var contentContainer = filterView.setupFilter(c);
         
-        // Begin code common to continuousFilterControl and dateFilterControl
-        // TODO: Consider extracting this to setupFilter
         var slider = contentContainer.append("div")
                 .classed("filter", true)
                 .classed("slider",true)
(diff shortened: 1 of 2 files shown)


docs/tool/js/views/filter-view.js
```diff
@@ -714,6 +706,188 @@ var filterView = {
         setState('nullsShown.' + c.source, ths.nullsShown);
 
     },
+    zoneAggregateFilterControl: function(component){
```
NealHumphrey: I'm concerned about the quantity of duplicated code in this approach. Is it necessary to create a whole new FilterControl element? 

I think the core difference between this FilterControl and the normal ContinuousFilterControl is that the data column that needs to be used depends on which zone type is currently selected. The min/max then will also need to change. Can we put those couple of changes into their own function or if statement and then use the existing code for the rest of it?

NealHumphrey:
Superseded by #451

Issue #388, opened 2017-07-11
Create ZoneFacts table and endpoints


NealHumphrey:
Instead of using our API endpoints to recalculate derivative fields adhoc, and instead of performing client-side reconfiguration of data into display fields, we want a master table to do these calculations every time new data is loaded and make it easier to join all these fields to the project table. We've previously referred to this as the 'NeighborhoodFacts' table. 

This ticket should a) create the table and b) make the data in the table part of the filter_data endpoint (which has a record for each project/subsidy combo). This will make all of this data available to the combined filter/layer view that is being used in V2. 

- [ ] Add a function that runs after all data loading is completed that drops the existing records from the zonefacts table and recalculates all the fields. Each row of the table should be one "zone" with the relevant calculated fields. Initial columns should be those that exist in our existing layers data set, with rows filled in for each zone for each of "ward", "neighborhood_cluster" and "census_tract". 
    - [ ] zone_type and zone_specific  (i.e. "ward" and "Ward 1" or "neighborhood_cluster" and "Cluster 32"). zone_type should match the relevant standard column name and zone_specific should match the value for that column that is used in other tables
    - [ ] Calculated fields from the "census_with_weighting" api endpoint: 'poverty_rate', 'fraction_black', 'income_per_capita', 'labor_participation', 'fraction_foreign', 'fraction_single_mothers'
    - [ ] All additional columns in the census table with weighting factors applied for each zone type (including duplicate rows for census_tract, which will completely duplicate the census table for ease of use). 
    - [ ] building permits: all permits past 3 months, construction permits past 3 months
    - [ ] crime: violent crime past 3 months, non-violent crime past 3 months, all crime past 3 months, and each of these columns as a rate per 100,000 people (copy code from "summarize observations" api endpoint).

- [ ] Left Join the project table to this table within the filter_data API endpoint. Do a left join 3 times, once each for ward, neighborhood_cluster and census_tract. For each join, the zones should be renamed `zonetype_columnname`. This means each row of the project data will have the denormalized data associated with all relevant zone types.

NealHumphrey:
Merging initial portions of this into dev per PR #416 

Still to do:
- Make changes as noted in the PR review

NealHumphrey:
Merged in some more changes in #436 . Changes requested from that code version:

PR changes requested by @nealhumphrey
[...]

PR #446, opened 2017-08-06
388 zone facts table

jkwening:
Finally finished completing unit test confirming validity of zone_facts calculated values. This should wrap up #388

@NealHumphrey is this task needed: complete get_residential_permits method that is called in summarize_observations method? It was a todo note from when I pulled in summarize observations code - currently calculating rate for building_permits is incomplete but not sure if there's an actual need for it.


NealHumphrey:
@jkwening thanks! I can review this tomorrow. 

On your question, yes we do need that. But we can make it a separate issue  - it's adding another field to the zone_facts table. I made a separate issue for it: https://github.com/codefordc/housing-insights/issues/447

PR #448, opened 2017-08-07
Refactor continuous filter

NealHumphrey:
Our filter components, and their links to the state module, have gotten complicated. The continuous filter is the most complicated, as it has text boxes, null checkboxes, sliders, and associated states for all of these. 

I'm trying to refactor this code to make it a bit easier to follow, in preparation for (manually) pulling in the code for #444. This PR is mostly rearranging the order of the code in the continuousfiltercontrol setup function so that code that is related to each other appears side by side, plus the addition of some code comments. Making this a separate PR so that the manual additions of #444 are clearly differentiated from this rearrangement.

### DIFF
--- a/docs/tool/js/core/housing-insights.js
+++ b/docs/tool/js/core/housing-insights.js
@@ -49,26 +49,42 @@ function StateModule() {
 
     function setState(key,value) { // making all state properties arrays so that previous value is held on to
                                    // current state to be accessed as state[key][0].
-        if ( state[key] === undefined ) {
+        var stateIsNew = (state[key] === undefined);
+
+        
+
+        if ( stateIsNew ) {
             state[key] = [value];
             PubSub.publish(key, value);
-            console.log('STATE CHANGE', key, value);
-        } else if ( state[key][0] !== value ) { // only for when `value` is a string or number. doesn't seem
-                                                // to cause an issue when value is an object, but it does duplicate
-                                                // the state. i.e. key[0] and key[1] will be equivalent. avoid that
-                                                // with logic before making the setState call.
-            state[key].unshift(value);
-            PubSub.publish(key, value);
-            console.log('STATE CHANGE', key, value);
-            if ( state[key].length > 2 ) {
-                state[key].length = 2;
-            }
+            console.log('STATE CREATED', key, value);
+
         } else {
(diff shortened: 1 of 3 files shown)


NealHumphrey:
Latest commit changes how setState behaves. It can now compare the state of arrays before updating the state (previously it only did this if the value was a string/boolean). This means we can set the state of a filterValues component multiple times without triggering a recalculation of the filteredData list - the redundant setStates will be ignored. Hopefully this will help us clean up some of the performance issues. 

@ostermanj I'm especially interested in your input on this approach, and if there was any additional reason you only did this for the boolean/string states.

ostermanj:
@NealHumphrey : the performance is much better! I knew the filterData was being recalculated over and over again and was especially bad on Clear All Filters, but wasn't sure why.

There was no good reason why comparing states only worked for strings and booleans except that they were the first states we had and we never extended it to handle other types. 

Happy to approve unless you want further review. I tested compatibility with the additions from my last PR and it all looks good.

NealHumphrey:
@ostermanj the one thing I wasn't sure about in this PR is that on initial page load I am getting a few of these errors:

```
PubSub.js:55 Uncaught Subscription token is already in use.
reThrowException @ PubSub.js:55
setTimeout (async)
callSubscriberWithDelayedExceptions @ PubSub.js:63
deliverMessage @ PubSub.js:82
deliverNamespaced @ PubSub.js:93
setTimeout (async)
publish @ PubSub.js:129
PubSub.publish @ PubSub.js:141
setState @ housing-insights.js:75
(anonymous) @ housing-insights.js:198
etc...
```

It appears to be assosciated with:
`STATE CHANGE dataLoaded.raw_project true`
(it's hard to be certain due to some async console logging)

Which is why it is only happening on page load not later on. I think this error was introduced in this PR because I don't recognize it. 

It doesn't seem to be affecting behavior at all, but if you are able to take a quick look or if you've seen that error before it would be useful to know if I've broken something. Otherwise we can track it down later as it doesn't seem to be breaking anything.

NealHumphrey:
Also, note that the many-recalculations when clicking 'clear all' is currently still occuring the first time you click it, but not later - because the unused filters have not been set in the state object yet, so the comparison turns up false. The array compare will allow us to fix this , but we'll need to initialize all the filters to empty arrays on page load. Can do that next.

Issue #442, opened 2017-07-31
Filter UI not reset properly in some circumstances


NealHumphrey:
- [x] Resetting values to original levels doesn't clear the filter pillbox. 
    - This works: uncheck 'unknown values' checkbox => pillbox added; recheck unknown values => pillbox removed
    - Doesn't work properly: 
        a) Slide value away from min or max; 
        b) slider value back to maximum/minimum
        Actual: pillbox not removed, filter values are simply set to their min/max. 
        Expected behavior: resetting the slider to a level matching its original level will set that filter as 'cleared'. 

- [x] Pressing 'x' in the pillbox toggles the 'unknown values' checkbox to the wrong state in some circumstances. This occurs if both the checkbox and the slider have been used. This is caused by the first problem in combination with some incorrect logic about checkbox state. 
    - Doesn't work properly:  For one of the slider data sets (e.g. Total units), 
        1) slide filter to some value 
        2) uncheck the checkbox 
        3) re-check the checkbox
        4) click the 'x' in the pillbox. 
        Actual: Checkbox is switched to unchecked when the filter is cleared. This does not accurately reflect the state of the filters, which in reality do include nulls. Clicking the checkbox again once does nothing, and clicking a final time restores the sync of state to ui. 
        Expected: Checkbox remains unchanged when clearing filter because it is already checked.

NealHumphrey:
#451 resolves the second issue. First issue still outstanding.

ostermanj:
First issue is addressed in PR #458

PR #451, opened 2017-08-08
392 zone filters + continuous and nulls refactor

NealHumphrey:
This pull request includes a variety of changes to the filterControl elements, primarily for continuous controls but also for date controls, and to the connected filter.js, router.js. Note that this PR also includes #448 which did some simple rearrangment of the continuous filter. Included:

- Any continuous control can have its min/max values reset at any point. 
- Zone-based continuous filters have their values cleared and min/max values reset whenever the zone type is changed. 
- Rearranges how values of the UI components (checkbox, textbox, slider) are bound together. If adjusting one of the UI components, the new state is determined by what is currently in the other UI components. This way if the internal state somehow gets out of sync with the UI state, it will self correct whenever a user uses any of the UI for that filter (see bugs resolved below)
- Stores all the UI components on the filterControl itself so they can be accessed by other code. 
- Removes the state of `nullsShown.<source>` in favor of using the 3rd value of the `filterValues.` state array (i.e. `[min,max,nullValue]`). The URL encoding code already added this to the state, which was duplicative, and we want the checkboxes to cause the same reactions in the UI as the sliders re: pillboxes etc. - @ptgott created the nullsShown state to separate the checkboxes from the code that executes filters, but that turned out to not be the desired behavior and maintaining two separate states was more complexity. 

This also fixes a few bugs that existed in the current dev branch:
- When loading a saved state, the checkbox for nulls could get unsynced from the values when includenulls was false, meaning the checkbox showed the opposite of what the state was. (found during dev, no issue created)
- Similar bug when executing filter updates in a certain order as described in bullet 2 of #442 

Additional bonus feature:
- Changes how the setState() command logs to the console. It now prints a grouped log that includes a stack trace, so coders can see which file initiated the new state for easier debugging.

### DIFF
--- a/docs/tool/js/core/housing-insights.js
+++ b/docs/tool/js/core/housing-insights.js
@@ -57,7 +57,7 @@ function StateModule() {
             state[key] = [value];
             PubSub.publish(key, value);
             console.log('STATE CREATED', key, value);
-            console.trace("Stack trace for state " + key)
+            console.trace("Stack trace for state " + key);
 
         } else {
             //If it's a string or array and values are the same, stateChanged=False+
@@ -76,14 +76,14 @@ function StateModule() {
                 PubSub.publish(key, value);
 
                 console.log('STATE CHANGE', key, value);
-                console.trace("Stack trace for state " + key)
+                console.trace("Stack trace for state " + key);
                 if ( state[key].length > 2 ) {
                     state[key].length = 2;
                 }
             } else {
                 //TODO do we want for there to be another 'announceState' method that publishes every time it fires, even if the value remains the same??
                 console.log("STATE UNCHANGED, NO PUB:", key, value);
-                console.trace("Stack trace for state " + key)
+                console.trace("Stack trace for state " + key);
             }
         }
         console.groupEnd()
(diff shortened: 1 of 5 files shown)


NealHumphrey:
Due to the fact that we're having a user test tonight at 8pm, I've merged this into the live website so that the zone-based filters will operate correctly for this test. This PR is into dev, so if there are any changes needed to the PR I'll work out the Git merging needed to get everything up to date when it's all resolved. 

@ptgott this PR is in conflict with your PR #444 and also undues some of your code on the nulls from a couple weeks ago.  I'm sorry for that, it's very soon for a refactor and I hope it's not too much of a dissapointment! There were a few conflicts and bugs in the existing system and it was easier for me to rearrange some of it rather than patch.

ostermanj:
Changing zone types doesn't seem to be clearing the zone-based filters the way we expect. @NealHumphrey

ostermanj:
@NealHumphrey : as discussed, I'll approve and merge this PR and write up the outstanding bugs as new issues so that other PRs are on top of this one.

PR #453, opened 2017-08-09
WIP -   437 building permit update

jkwening:
## What's new

- Fixed bug causing --update-only to fail in LoadData.py

- Added cleaner step that geocodes the census_tract of building permits.

## Notes
Due to size of building permits data needed to improve processing time. For 2016 data file, there's 47k rows with ~37k unique buildings. It was taking 45+ to process and that's if it didn't time out  - I was able to process without issues during evening hours.

#### Solution
Implemented concurrency since the bottleneck was an IO issue - api calls and we don't care the order in which we write the data rows to the psv file for database upload. 

It is currently using a max of 100 threads to process and clean all the rows for a raw data file. This improved processing time considerably - went from 45+ mins to 5-7 mins (398s last night, 329s early this morning).

## Things To Consider: 

- Using threading did cause one issue - too many api calls at times for the same value (mar_id). Resolved this by adding a while loop that waits a second and then tries again when an Exception is thrown. This works nicely - 1s wait seemed the optimal solution; 0.5s and 0.75s actually took much longer to complete but that could be due to me testing after 8am.

- I think we should move this up into the batch api raw data download process and before cleaning process and database loading. I think this would allows us to continue to get the benefit of improved processing time and yet minimize the timing issue when we manually refresh the database when user is testing something out locally. I know others have been having issues updating locally due to mar api timing out at times.
  - I'm thinking add a procedure that is called for any api data that will need geocode cleaning. It will query and update any mar_id location that isn't currently available. So we will end up with mar_geocode table with complete geocode data for most/all buildings - maybe skip if times out and attempt again when actually loading raw data into database: undecided on this right now. The cleaner will lookup using the table and only make api call if building geocode info is missing.
  - I know it seems we're duplicating the work here but this front loads most of the work associated with mar api calls and only makes calls later if the table is missing information and will update the mar_geocode table accordingly. So with each iteration, less calls will be made until the entire batch process is done again.
  - Also, we're already doing some mar api calls in batch processing already so I should be able to connect this all together for smoother processing in general with less requests make to mar api.

## Testing
Run LoadData.py with `docker --update-only building_permits_2016` if database is already built or rebuild the entire database with just `docker`.  **Note: I recommend attempting after 5pm or running on something smaller, maybe project or dchousing_project.**

## To Do

- Process and load building_permits_2017 data. Note - this raw data is still causing something to error out - investigating further to see if I can figure out what is going on.

### DIFF
--- a/python/housinginsights/ingestion/Cleaners.py
+++ b/python/housinginsights/ingestion/Cleaners.py
@@ -66,7 +66,7 @@ class CleanerBase(object, metaclass=ABCMeta):
                 row[key] = self.null_value
         return row
 
-    def remove_line_breaks(self,row):
+    def remove_line_breaks(self, row):
         #TODO see if it's possible to not do this by getting the copy_from to be ok with breaks
         for key in row:
             row[key] = row[key].replace('\r','__')
@@ -88,7 +88,7 @@ class CleanerBase(object, metaclass=ABCMeta):
 
         return date
 
-    def convert_boolean(self,value):
+    def convert_boolean(self, value):
         mapping = {
             'Yes': True,
             'No': False,
@@ -115,7 +115,7 @@ class CleanerBase(object, metaclass=ABCMeta):
             row[source_name] = self.format_date(row[source_name])
         return row
 
-    def remove_non_dc_tracts(self,row,column_name):
+    def remove_non_dc_tracts(self, row, column_name):
         '''
         TODO change to use self.census_mapping
         '''
@@ -144,7 +144,7 @@ class CleanerBase(object, metaclass=ABCMeta):
         else:
             return None
 
-    def rename_census_tract(self,row,row_num=None,column_name='census_tract'):
+    def rename_census_tract(self, row, column_name='census_tract'):
             '''
             Make all census tract names follow a consistent format. 
             column_name corresponds to the key of the row, which depends on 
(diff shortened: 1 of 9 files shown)


NealHumphrey:
@jkwening great stuff! 

After your question last night and similar 'Notes' here, I looked around a bit to see if there was another solution. I found that opendata.dc.gov has the [Address Points data set](http://opendata.dc.gov/datasets/address-points) which appears to be a direct extract of the MAR. It has another data set for intersections, which I think comprises the rest of the MAR. Since we have the mar_id already we can look it up directly from this table. 

What do you think about:
1) Ingest this data set as a new table in the database
2) Keep all your current code that looks up from the MAR API as the fallback when checking our internal database doesn't return a result
3) Add a step 1 to `add_census_tract_from_mar` that cheks the internal database first
4) We could also do this for geocoding of the projects - though in that case we don't have the mar_id yet and we will have to match the full address itself. Since the MAR API provides some fuzzy-matching logic, we will likely still need the MAR API in many cases, but could speed things up by getting what we can from the opendata data set first.

jkwening:
Thanks. Yeah I was going to see if I could incorporate that into the solution as the base data for the mar_geocode table and build from there. I added Address Point data set into the api batch code when I was working on the current cleaner code used for the project table.

Your outline mirrors what I was thinking, so start heading in that direction. Thanks.

PR #458, opened 2017-08-09
Jo 442 filter ui bugs

ostermanj:
Addresses first bug listed in issue #442 .

Pillboxes and URL encoding clear when date and continuous filters are set back to their original settings. Works on sliders, checkbox, and text inputs.


NealHumphrey:
Works great!

PR #462, opened 2017-08-12
Remove table during update

NealHumphrey:
Provides 2 ways for a table to be dropped (along w/ associated sql manifest rows) when updating the database. This allows for the addition/removal of columns while updating a single table, instead of only being able to do so by rebuilding the whole database. 

Also includes a few try/except blocks for the zone_facts table. Add logging.error when a field could not be found/calculated due to missing data - this means the rebuilding of the database will still work even if a necessary table (building_permits, census, etc. ) is missing. Allows for faster testing of things that only apply to certain sections of the database b/c you can add 'skips' to any tables you're not currently working on.

### DIFF
--- a/python/housinginsights/ingestion/LoadData.py
+++ b/python/housinginsights/ingestion/LoadData.py
@@ -27,6 +27,7 @@ from sqlalchemy import Table, Column, Integer, String, MetaData, Numeric
 from datetime import datetime
 from uuid import uuid4
 import dateutil.parser as dateparser
+from sqlalchemy.exc import ProgrammingError
 
 # Needed to make relative package imports when running this file as a script
 # (i.e. for testing purposes).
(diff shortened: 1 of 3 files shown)


NealHumphrey:
To test: 

Before running the command, check the 'crime' table and sql manifest table - there should be 3 crime data sets: 2015, 2016 and 2017. 

From the /python/scripts folder, run this command:
`python load_data.py docker --remove-tables crime --update-only crime_2016 crime_2017`

What should happen:
This will first drop the crime table, and then update the database with the 2016 and 2017 crime data. 
The manifest sql table will now only have two crime entries, matching these two fields, as will the crime data set. 

To restore your database to original state by adding in the missing data set:
`python load_data.py docker --update-only crime_2015`

Issue #333, opened 2017-06-12
'clear all filters' triggering multiple state changes


NealHumphrey:
The "Clear All Filters" button introduced in #324 has some performance issues where it is triggering multiple state change events, causing the page to perform the filtering calculation multiple times. 

When the 'clear filters' button is pressed, there is a state change of filterValues.data_id for:
1) 2X for every slider filter (presumably once for min and once for max?)
2) for every categorical filter that had been selected

Therefore, the 'filteredData' event occurs repeatedly for every one of these. The problem is especially acute for the sliders, since it occurs even if no sliders had been selected. 

Best guess of cause:
- Slider callbacks use the nouislider 'set' event, while the categorical event uses the 'change' event
- clearAllFilters method calls the 'clear' method on every single filterUI, not just those that have been selected. 

Possible solutions:
- Partial solution is to have clearAllFilters cycle through the filterValues state instead of the filterControls list. This will only trigger it once for every that is actually being cleared. 
- More thorough solution would change how the `['filterValues', filterUtil.publishFilteredData]` subscription works, so that if a user changes something the filter is triggered, but if some code triggers the filtering it  only does it at the very end after all filterValues have been adjusted to their desired state. We should be careful to keep the logic simple instead of doing a patch solution that adds additional layers of logic. Probably worth discussing best approach on this one.

ptgott:
This all sounds reasonable! We could add an 'is_touched' method to each filterControl that detects non-default values. This way we would cycle through only those filterControls that have been 'touched' when calling clearAllFilters.

It'd be great to decouple the setState(filterValues, state) from the user's adjustments of the filterControls, so that changing the filter control doesn't automatically change filterValues. As of now, changing filterValues directly leaves the controls themselves unchanged, and changing the controls after clearing filterValues just changes filterValues again due to the 'set' or 'change' handlers.

We could change the event handlers/callbacks for the filterControls so that rather than call setState with filterValues, they set a state related to the controls themselves. For example: setState('filterControlChange, { name: <data name>, values: <values> }. This way, we could be more explicit about what responds to a change in the filterControls, whether it's an individual filter value or some other object (one that would clear filters in a batch). I'm not sure yet how exactly this would work, though!

ptgott:
I've prepared a quick fix in #367 that adds an 'isTouched' method to the filter controls based on whether their current values match their initial ones, then uses this method to decide which filters to clear in clearAllFilters. 

[...]

PR #464, opened 2017-08-12
333 Fix ContinuousFilterControl.isTouched() and dateFilterControl.isTouched()

ptgott:
Addresses #333 

The former had tried to match the string return value of the text inputs with the numeric return value of min/maxDatum.

The latter had tried to use an equivalence operator on JS Date objects, which can't be done. I've used the operator on the valueOf() those Dates instead.

Now only 'touched' filter controls should clear.


NealHumphrey:
Great!

PR #465, opened 2017-08-13
Dedupe project and DHCD datasets

NealHumphrey:
This is a first pass at implementing the matching of records in the DHCD dataset to those that already exist in the project table. It uses the same method as the DCHousing dataset to create project and subsidy tables out of a single "project" table. Currently just using simple string matching against the `mar` table of the database to geocode the rows of the DHCD data set. 

Still to do:
- Implement better geocoding w/ calls to the MAR api 
- Deal with dirty address data - multiple addresses in one line, sometimes written out sometimes just the street numbers
- Use the 'dhcd projects' table (equivalent to our subsidies table) to better describe subsidies. Currently this file is coming in blank from the API requests. 
- Map more fields

Note, this PR also includes the changes in #462 

## To test
- Add the mar data set to your database:
    - edit get_api_data.py to use `unique_data_ids = ['mar']`
    - from /python/scripts run `python get_api_data.py` to download a local copy of the mar data, which should update your manifest.csv
    - run `python load_data.py --update-only mar` to add the mar to your database

- Get the DHCD data
    - edit get_api_data.py to use `unique_data_ids = None` and `module_list = ['dhcd']`
    - run `python get_api_data.py`

### DIFF
--- a/python/housinginsights/ingestion/LoadData.py
+++ b/python/housinginsights/ingestion/LoadData.py
@@ -27,6 +27,7 @@ from sqlalchemy import Table, Column, Integer, String, MetaData, Numeric
 from datetime import datetime
 from uuid import uuid4
 import dateutil.parser as dateparser
+from sqlalchemy.exc import ProgrammingError
 
 # Needed to make relative package imports when running this file as a script
 # (i.e. for testing purposes).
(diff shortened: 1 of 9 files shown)


NealHumphrey:
Added another commit - the code now checks the MarApi if it doesn't find a match in the mar database directly.

NealHumphrey:
This is our current performance:
* About half of the properties in DHCD's database are matched to an existing record in the project table. 
* Nearly all of DHCD's properties are matched to a MAR id. From a quick spot check, almost all of the unmatched addresses are for DHCD address fields that use a street number range instead of a single street number, i.e. "2720-2734 Langston Lane SE". We should be able to snag the rest of these with a regex that swaps these out for either end of the range. 
* We're still just taking the first address if multiple addresses are provided. We might be able to bump of the half that are matched in the proejct table to find some additional duplicates by matching to other addresses listed in the address field of DHCD's data. In additional, a final pass could be looking for a similar project name. Some manual analysis could tell us about how many we think we're currently missing.

jkwening:
@NealHumphrey regarding the unmatched MAR ids, this is a consistent issue I've been running into for most ingested data sets, specifically project and building_permits. What I've found is that the raw data set actually has incorrect mar_id value. When I do a reverse look up using any of the other available geocode info I end up finding another mar_id with complete geocode data for the location. I'm going to create a ticket to create a reusable function that handles this better and avoid having to duplicate this task across the code base - I think I currently have similar code in at least two places.

NealHumphrey:
@jkwening I put my code for this into a function located in /tools, so we
can pull it into other places from there. In fact if you haven't added your
look up building permit from database Mar table code, we should use my
function for this too.

On Aug 13, 2017 1:20 PM, "John Kwening" <notifications@github.com> wrote:

@NealHumphrey <https://github.com/nealhumphrey> regarding the unmatched MAR
ids, this is a consistent issue I've been running into for most ingested
data sets, specifically project and building_permits. What I've found is
that the raw data set actually has incorrect mar_id value. When I do a
reverse look up using any of the other available geocode info I end up
finding another mar_id with complete geocode data for the location. I'm
going to create a ticket to create a reusable function that handles this
better and avoid having to duplicate this task across the code base - I
think I currently have similar code in at least two places.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/codefordc/housing-insights/pull/465#issuecomment-322054668>,
or mute the thread

NealHumphrey:
@jkwening If you haven't reviewed this PR yet, you might want to jump straight to  #466 - it adds the next step of actually loading the dhcd data set into the database after it's been generated, but while I was doing that I found a couple minor bugs that needed fixing in this code.

PR #466, opened 2017-08-14
Update project table

NealHumphrey:
This branch contains code fixes needed to get all 3 project data sources loaded into the database. 
This builds off of the PR #465 and includes all that code too. It also fixes a few bugs with that code. 

It also switches the manifest.csv to use the new S3 bucket from the Code for DC AWS account, instead of the one from Neal's personal account that we had been using. 

To test:
`python load_data.py docker --remove-tables project --update-only project prescat_project prescat_subsidy dchousing_project dchousing_subsidy dhcd_dfd_properties_project dhcd_dfd_properties_subsidy`


NealHumphrey:
Hmm, so it looks like `mar_api.find_location()` returns a different format than the `.reverse_lat_lon` and `.reverse_geocode()` methods, due to the MAR api returning a different result format. I lifted the that set of conditionals from my other code that only used `.find_location()`. 

This also means that this line `row['mar_id'] = result['Table1'][0]["ADDRESS_ID"]` would never have worked if the `.find_location()` method were implemented - I think it just never got called during the DCHousing data set since we always had lat/lon available. 

For now I'll hotfix this by putting the `return` method underneath each conditional. But, we'll need to consider broader solutions as part of our enhanced dedupe code - and we might want to consider adding some wrapper methods to the `MarApiConn` that return a consistent format and handle things like finding intersections instead of addresses.

NealHumphrey:
Made change; testing and it works on my machine. Going to merge.

PR #473, opened 2017-08-18
472 search for project

NealHumphrey:
Adds a searchbar to the top navigation, and adds two filter categories of project name and project address to the left sidebar. The searchbar uses a combined name/address field called 'proj_name_addre' which has been added to the API for the /filter and /projects endpoints. This has already been deployed to the live API so that this functionality can be tested w/out spinning up a local api server and swapping urls. 

The search bar is a re-styled combo search/dropdown box, the same one we use for categorical data choices, and behaves like a regular filter control - this seemed like the fastest way to get a basic search functionality integrated into our current workflow. One downside is that it doesn't produce true 'search results' like a user might expect, rather it provides instant filtered results in the dropdown menu. 

Currently this simply applies the filter the way a regular filter works, i.e. highlighting the project on the map. However, we could attach another event to the searchbar selection that takes them to the project view associated with the one matching project. Not sure which behaviour users would find more intuitive. At a minimum, we could activate the 'matching projects' tab, since the 'summary' tab does not make much sense when only one building is selected. 

Should be ready to merge. Note the encoding/decoding on the url is working, though there seems to be a bug in all url decoding that causes it to fail to color code the dots sometimes.

### DIFF
--- a/docs/tool/css/styles.css
+++ b/docs/tool/css/styles.css
@@ -710,3 +710,22 @@ dl.inline dt{
   display: inline-block;
   min-width: 75px;
 }
+
+
+/*search bar*/
+#searchbar .dropdown {
+    border-radius:15px;
+    max-width:400px;
+    float:right;
+    margin-right:10px;
+    margin-left:50px;
+}
+
+#searchbar {
+    height:30px;
+}
+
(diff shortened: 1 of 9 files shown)


NealHumphrey:
Approving my own PR in the interest of getting our code base caught up

PR #475, opened 2017-08-18
Adds nested JSON data for one-to-many tables in project endpoint

NealHumphrey:
hat currently have an nlihc_id. Currently this means that the `topa`, `subsidy`, `real_property` and `reac_score` tables are now available on the project endpoint whenver you request the data for a specific nlihc_id. For performance reasons the data from these tables are not added to the json when a list of all projects is requested.

- Adds sql-injection protection to this endpoint

To test:
- Spin up server with `python application.py`
- visit localhost:5000/api/projects/NL000196

Deployed to live server

### DIFF
--- a/python/api/project_extended_constructor.py
+++ b/python/api/project_extended_constructor.py
@@ -2,7 +2,9 @@ from flask import Blueprint
 from api.utils import get_zone_facts_select_columns
 from flask import jsonify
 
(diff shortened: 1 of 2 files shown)


NealHumphrey:
Deployed to live server to make available for front end devs, mergeing

Issue #457, opened 2017-08-09
Extend URL encoding/decoding to handle location filters


ostermanj:

ostermanj:
address in PR #478

NealHumphrey:
Verified this works in #478, closing - thanks John!

Issue #468, opened 2017-08-15
Clean up scroll behavior in data choices list plus make dataChoices behavior globally accessible


NealHumphrey:
In the data choices list on the sidebar, selecting a new data choice collapses the open data control and opens the selected control (due to magic of semantic ui's accordion). However, if the data choice is not near the top of the list, it can appear open below the current scroll level, causing the user to have to manually scroll down. 

- [ ] in the clickCallback on row 775, use setState to announce the selected data source i.e. `setState('dataChoice',c.source)`. This will also give the URL encoding something to tap into for saving which dataChoice was open in the view. 

- [ ] Refactor the clickCallback into a globally accessible function `selectDataChoice` so that it can be called from elsewhere (this sets us up for url decoding, since we're working on this anyways). 
- [x] Auto-scroll the sidebar so that the selected data choice is at the top of the visible pane when that function is run. 
- [x] Add extra padding of 150 pixels below the `#filter-components` div. This combined with autoscroll will solve the scroll item in #396

NealHumphrey:
You can test the issue in #396 by moving the 'portfolio' choice to the bottom of the list in `dataChoices.js` (they are rendered in order). the dropdown appears out of view.

ostermanj:
I wonder if we should allow more than one accordion to be open at a time. I could see that be more expected and certainly gives more control to the users, but if no one's mentioned it perhaps it's not an issue. It would be nice when decoding filter settings from a url to show the active filters, for instance. It wold affect the behavior of showing  data overlays, which would have to activate only on clicks that open the associated according. Currently activates on click. Thoughts?

ostermanj:
I've addressed this issue in later commits of PR #478 . I included it there to avoid conflicting with it. The mechanisms are a little different than laid out in the checklist above, with the main difference being that only overLay dataChoices are being url encoded and decoded. That functionality could be extended to non-overLay dataChoices, but unless their filters are activated there doesn't seem to be reason to do so, and any activated filters are already being encoded.

Given that, I'm using the existing `setState(overLay, ...)` to trigger the state encoding rather than creating a new `setState(dataChoice, ...)`.

The auto scroll is handled via an onOpen callback method native to the semantic ui accordions rather than in the clickCallback.

The overLay decoding programmatically 'clicks' the associated accordion element.

NealHumphrey:
[...]

Issue #467, opened 2017-08-15
Extend url encoding/decoding to handle decimal values


NealHumphrey:
Any of the dataChoices that are formatted as a 'percent' (e. g. poverty rate) have a value of less than 1. URL encoding currently rounds this to 1 or zero. Should instead allow these to have 2 decimals of resolution, i.e. 0.03 or 0.54.

One option: To keep url short, could format anything that is a percent data field as multipled by 100? Then the url will be x=23-56 for the field x with values of 0.23 and 0.56.

ostermanj:
Addressed in PR #478. Any needed rounding is happening upstream in the filter controls, so just removing the extra rounding in router.js seems to do it.

NealHumphrey:
@ostermanj This works on the encode portion, but decoding does not seem to work. If I set a filter (e.g. poverty rate) the url changes. If I then refresh the page to reload it with the url that now has poverty rate encoded in it, instead of applying the filter the url decoder erases the rate data from the url and loads the page without that filter applied. If there are other filters encoded in the url, those filters are not affected and are appropriately applied. 

I will not prevent this from merging #478 since the first part is working and is an improvement to current status, but I will keep this issue open so we can resolve this in a future PR.

ostermanj:
Ok -- I'll check on it. Thought it was working . . .

ostermanj:
The decimal filter is working for me on Chrome, FF, and Safari (Mac). I wonder if it's a cross-browser / platform or edge-case issue. @NealHumphrey can you still reproduce the error?

NealHumphrey:
@ostermanj I couldn't recreate. Assuming it is fixed now so closing ticket

PR #478, opened 2017-08-19
Jo 457 url encode location

ostermanj:
Adds locations filters to the url encoding/decoding functionality. Also encodes/decodes mapLayer selections other than the initial setting, mostly so that decoding the location from a non-initial (ie non-Ward) zone type will work.

Re: issue #457

### DIFF
--- a/docs/tool/css/filters.css
+++ b/docs/tool/css/filters.css
@@ -9,6 +9,12 @@
 
 
 /*Stuff for all filter components*/
+
+#filter-components {
+  padding-bottom: 150px;
+  
+}
+
 .filter-group {
     border-bottom: 1px solid black;
 }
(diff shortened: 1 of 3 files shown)


ostermanj:
I'll add #467 to this . (decimal values)

ostermanj:
About to add commits addressing issue #468 re: scroll behavior and encoding overLays. I'm including in this PR to avoid merge conflicts.

NealHumphrey:
Approved, merged manually via command line to resolve conflicts between this and the searchbar. 

Note, the decoding of percent-based sources is still outstanding as described in #467 

In addition, I moved the .accordion settings outside of the for loop in the `buildFilterComponents` function - doesn't need to be reset every time; it's working in its new home.

Issue #384, opened 2017-07-06
Onboarding instructions are out of date


jasonrhaas:
The commands to load data in the database have changed slightly.  Need to update the onboarding instructions to reflect this.

http://localhost:4000/resources/onboarding/docker.html

instead of `python load_data.py docker rebuild`, 

its now `python load_data.py docker` or `python load_data.py docker -s`

PR #480, opened 2017-08-22
fix docker onboard to remove rebuild

citizenrich:
Fixes the instructions for loading data using Docker.

Addresses https://github.com/codefordc/housing-insights/issues/384


NealHumphrey:
Great

PR #482, opened 2017-08-23
cleaned up documentation for projects table

emkap01:
@NealHumphrey this is a first crack at cleaning up the documentation for the projects table. All I did was update the description of the different datasets that we use to build the table, since the Markdown file was out of date by several months.  

I added context per dataset based on whatever notes I could find in the relevant Issues, and also removed all hyperlinks to individual Issues (on the assumption that when users of the tool actually call up the documentation from the front end, they will neither need nor want to see the underlying Issues).

It would probably be worthwhile to add more language in here that explains the business logic behind how we dedup projects across datasets, but I will defer to you and/or @jkwening for that since a) much of that logic is baked deep down into the code; and b) I'm not sure how much of it may yet change going forward anyway.

### DIFF
--- a/docs/data/affordable-housing-inventory.md
+++ b/docs/data/affordable-housing-inventory.md
@@ -4,42 +4,39 @@ tablename: project
 title: Affordable Housing Inventory
 ---
 
-**NOTE** This page is incomplete, but will document the source of data about buildings that are shown in the Housing Insights tool. More thorough documentation will be ready for our August release. 
-
-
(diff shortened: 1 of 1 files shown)


NealHumphrey:
Thanks @emkap01 ! We will want to add the business logic description, but it is in flux right now - @jkwening is looking into one-to-many addresses, and we're adding some additional validation steps. I'll work on writing that up once we've stabilized that.

Issue #476, opened 2017-08-18
Small mapview edits per user feedback


NealHumphrey:
Will add to this issue as I compile notes from the 3 recent user tests. Some bigger changes from user tests are already in their own issues (i.e. search box, nulls)

- [x] Map view zooms too close - add more padding and a max zoom level
- [x] Make link to the specific project page more obvious (put a 'see more' link in map tooltip, put a 'more info' button in the preview.)
- [ ] Add list of subsidies with start/end dates to the project preview 

more to come

PR #488, opened 2017-08-25
Made preview link more obvious, updated null matching

jmithani:
Addresses #476.

In the preview pane, makes the link more obvious by adding '>>' and underlining. 

For showProjectPreview, updated the display of null values by adding 
`| value == "null"`. Not sure how pervasive matching on `null` is versus `"null"` -- something to maybe check for in the future?

Thanks for introducing me to the project and helping me contribute!


NealHumphrey:
Thanks @jmithani ! I tracked down the source of the 'null' bug you found and made an issue to treat it at its source, so thanks for point that out as well!

PR #491, opened 2017-08-25
437 building permit update

NealHumphrey:
@jkwening - this branch is the same as your 437 branch, with the addition of another commit I added. 

Speeds up the building permit geocoding by checking the internal database copy of the MAR:
- Loads the full mar table of mar_id, census_tract into memory as a dictionary
- clean(row) checks memory first, then queries database directly, then looks up via mar API, returning whichever geocoded row comes first.


NealHumphrey:
@jkwening I went ahead and added a similar method for geocoding all the other stuff we get from the MAR with the intent of using it to speed up the project table geocoding (i.e. dchousing). Unfortunately the database copy of the MAR doesn't have all the fields we need (streetview) so it doesnh't help us. 

I figured save the method for now, though, since I expect there are likely a few other things we might want to mar geocode in a similar way that won't need street view. TBD.

Issue #490, opened 2017-08-25
Navigation of active filters improvements


NealHumphrey:
a) restore the dark-orange highlighting for activated filters - the accordion header in left sidebar
b) make the pillboxes clickable so that if you click the pillbox (not the x) it will open that filter

History/Description copied from #468 discussion:
In v1, I also had activated filters change their dropdown header band color to be dark orange if they were activated, while they are the (current) light orange if they are not in use. Users in round 1 of testing didn't really notice that so I removed it, but I think that with the new behavior of the dark orange pillboxes rising from their source location it might now be clearer to users - and as you mention, a good visual cue when loading pre-filtered data from a url since hte user won't have past behavior to tell them what had been selected.

NealHumphrey:
Added in #492, closing

PR #492, opened 2017-08-26
Jo 490 filter nav improvements

ostermanj:
RE issue #490 : Active filters are darker orange again; clicking pill boxes scrolls to associated filter and opens them if closed.


NealHumphrey:
@ostermanj looks great! Thanks for adding this. Merging in.

PR #494, opened 2017-08-26
WIP - Address table: one to many address

jkwening:
## What's new

Added `get_unique_addresses_from_str` method in functions.py module.

The function takes an address string, parses the string, and returns a list of all
unique addresses based on the following delimiters:
    

- semicolon (;) - treated as delimiter for multiple unique addresses
-  'and' - treated as delimiter for multiple unique addresses
-  ampersand (&) - treated as delimiter for address number ranges
-  dash (-) - treated as delimiter for address number ranges

The function handles all potential composites of the delimiters noted above in a given address string. For processing composites, first handles 'complete' address delimiters (;, 'and'), and then address number range delimiters (&, -).

## Notes
I intentionally did not include any address validation in this method because I feel that should be decoupled and applied as need in context outside of this utility method.

## Testing
Review cases in unit test, run to confirm all tests pass. Add additional test cases if I missed any edge cases and run unit test.

## To Do

- Utilize the new function to complete the 'address table'

### DIFF
--- a/python/housinginsights/ingestion/functions.py
+++ b/python/housinginsights/ingestion/functions.py
@@ -138,6 +138,63 @@ def join_paths(pieces=[]):
     '''
     return '/'.join(s.strip('/') for s in pieces)
 
+
+def identify_valid_unique_address(address_str=""):
+
+    def _trim_str(add_str):
+        """Helper function that does some simple string cleaning."""
+        add_str = add_str.lstrip()
+        add_str = add_str.rstrip()
+        return add_str
+
+    result = list()
+
+    # 1: identify all full address delimiters - ';', 'and'
+    temp_results = address_str.split(sep=';')
+    if len(temp_results) > 1:
+        for address in temp_results:
+            address = _trim_str(address)
+            result.append(address)
+
+    temp_results = address_str.split(sep='and')
+    if len(temp_results) > 1:
+        for address in temp_results:
+            address = _trim_str(address)
+            result.append(address)
+
+    # 2: identify address number range delimiters  - '&', '-'
+    temp_results = address_str.split(sep='&')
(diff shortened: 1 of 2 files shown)


NealHumphrey:
@jkwening this is great, and especially thank you for the super thorough testing. 

We may end up needing a few more situations - unfortunately there's a bunch more messy formats. However I'd also like to send some of the one-off problem cases back up to the source and see if they'll fix them in their database instead of adding lots of extra logic to deal with it all on our end. We might consider adding a hotfix list for situations like that. Some examples in the DHCD data set that look like they might not be handled in this (albeit already robust) solution:

- Incomplete use of dash address: `1401-09 11TH ST NW`
- Comma separated numbers: `1825,1827,1829,1837,1839,1841,1843,1845,1860,1862,1864 CENTRAL PLACE, NE`
- Mixing of commas and & for numbers: `24, 52, & 230 BATES ST NW`
- Mixing of commas and dash separated numbers with &: `1416-1428,1432,1436 & 1440 R ST NW`
- Different meaning of  &: `HAYES STREET &ANACOSTIA AVENUE, NORTH EAST` (this would be falsely split into two addresses, even though it's actually a street corner that should be one invalid address)

The one of those that I think makes sense to try to add is the comma separated numbers. The other odd ones I might put into a hotfix list that we map from current messy format into a parseable format and put the hotfix mapping as the first step of the get address function. What do you think?

NealHumphrey:
Regardless of how we handle the rest of those cases, I'm merging this in now and will work on the next step of using this function to make the new table.

NealHumphrey:
Moved this into an issue so let's continue the conversation in #496 instead of this PR since after merging it is hidden from the main pull request list.

PR #497, opened 2017-08-27
WIP: Make proj-addre one-to-many data and load into database. 

NealHumphrey:
- Adds an "apiconn" object for the prescat. Currently this just loads the file from disk, but in the future this can be the class that checks for newly uploaded data in an online folder that admins can put data into
- Adds a function to the base_project.py apiconn that uses the split address function to create a csv file of the one-to-many project to addresses.
- Moves the split address function out of ingestion and into tools, since it's not used only in ingestion
- Fixes bug in the MarApiConn - adds a method to query api using address string instead of generic string. Helps eliminate false positives.
- Changes references to find_location to instead be find_addr_string to eliminate false positives
- Fixes bug in address splitting code that found `and` in `Marland` or `Randolph`. Changed split string to `' and '`.

- Adds proj_addre table to database
- DHCD and DCHousing ApiConn methods also create `_addre.csv` files for upload
- Entity resolution now done via the one to many mapping of addresses - if any of the addresses match, assumes the whole project matches. 

## To test
- Create the prescat_addre.csv file: from the `python/housinginsights/sources/` folder run `python prescat.py` (needs to be added to the options list in get_api_data.py, but should be ready to work there)
- Add the table, from `/python/scripts`: `python load_data.py docker --update-only proj_addre`
- Get DHCD and DCHousing data sets: `python get_api_data.py` with the code options edited to request the `DCHousing` and `dhcd` modules.


NealHumphrey:
Note - work in progress pull request. Making now just to share status and avoid conflicts; plan to pick up where I left off Sunday AM.

NealHumphrey:
Updated description to reflect progress w/ most recent commit

PR #498, opened 2017-08-27
adds parcel data set to database

jkwening:
Added parcel data to ingestion code.

Note - data for NL000414 uses "U" for parcel_x/y values instead of "N" for null. We should probably have this cleaned up in future updates from prescat.


NealHumphrey:
@jkwening re: the "U" for nulls, you can add it to the list of null values in the Cleaner for that table, like this:

`row = self.replace_nulls(row, null_values=['U', '','Null',None])`

PR #503, opened 2017-08-29
WIP: Jo ui suggestions

ostermanj:
This PR suggests change the UI display of how filters are scrolled and how the pillboxes appear. For review.

Instead of the entire sidebar scroll, in effect moving the zone-type selectors and the active-filter pillboxes out of view, only the inner container scrolls:

scroll

Zonetype descriptions are title text that appear on hover rather than static, where they took up valuable real estate.

Pillboxes fly to a position over the map but only appear when that area is hovered over. We had similar positions in earlier version and opted against it, but I think other changes made since then make this preferable:

flyto

Opening filter accordions or clicking on their pillboxes scrolls the filter to the top of the container:

open

### DIFF
--- a/docs/tool/css/filters.css
+++ b/docs/tool/css/filters.css
@@ -12,7 +12,9 @@
 
 #filter-components {
   padding-bottom: 150px;
-  
+  display: flex;
+  flex-direction: column;
+  overflow-y: auto;
 }
 
 .filter-group {
--- a/docs/tool/css/styles.css
+++ b/docs/tool/css/styles.css
@@ -199,11 +199,16 @@ button.nullValuesToggleSubmit::before {
     display: block;
     left: 100%;
     height: 40px;
+    width: 200px;
 
 }
 
 #clear-pillbox-holder:hover {
     height: initial;
+}
+
+#clear-pillbox-holder .ui.label {
+    min-height: 30px;   
 }
 
 .ui.label.clear-single {
@@ -586,6 +591,11 @@ a.active {
     font-weight: bold;
     border-top: 1px solid rgba(34, 36, 38, 0.15);
 }
+
+#filters {
+    overflow-y: hidden;
+}
+
 /*Mimics accordion content styling*/
 #filters .static-content {
     margin: 0;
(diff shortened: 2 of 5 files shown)


NealHumphrey:
@ostermanj 

I definitely agree with the zone type descriptors and the in-data scroll (though on Firefox this is not working as it does on Chrome - the right sidebar works properly in both so might want to compare the css). 

On the filter pillboxes, all of our users had some amount of learning barrier to understanding what a filter was, and often had a short memory span for when they had selected a filter - if they had an old filter still activated, they often forgot that it was still there later in testing when they were working with a different filter. While I think your changes definitely clean up the view, and as a power-user like it, I'm concerned that it might make things a little harder for users by obscuring which filters are active.  We have made multiple other changes to that behavior, though (fly to pillboxes, darker labels when active, etc.) so it's hard to know how they'll all balance out. 

I think maybe with a few more adjustments we could keep going in this direction of a cleaner look but still keep it obvious to users when they've forgotten about an old filter. @txomashernandez you saw all the user tests, do you have any thoughts on this as well?

NealHumphrey:
@ostermanj Been thinking a bit more about this. I think they key thing that's needed is for users to be shown what filters are active when they might have forgotten - i.e. when a new filter is selected. What do you think about the following tweaks to your PR as a potential way to do that:
 1) When selecting a new filter, the flydown menu is activated (as if the user had hovered), and then fades away ~0.5-1 second after the new pillbox arrives
2) The pillbox that is flying in goes into it's appropriate spot on the list instead of dissappearing into the general area of the 'filters' pillbox (this might happen naturally w/ #1)
3) On hover-out, add a slight delay to the collapse action on the pillboxes, so that users don't need to be quite as precise about their mouse actions when hovering over and down. 

I think that might be enough - what do you think?

ostermanj:
@NealHumphrey Yeah, it would be good to remind people that filters are activated and which ones without them having to hover over. I like it. Hopefully not too much additional to implement . . . I'll try to work on it today.

ostermanj:
@NealHumphrey New commits address your suggestions. The hover effects are via CSS, so I can't add a delay to the collapse without refactoring it to be javascript-based. I added bottom-margin to the last pillbox; that should make it so users don't have to be too precise with their mouse actions.

Can switch to mouseenter, mouseleave if it needs it, but thought I'd give that a shot first.

NealHumphrey:
Looks good - merging in.

Issue #499, opened 2017-08-27
Additions to export functionality


NealHumphrey:
- [x] Refactor export functions to be reusable
    - Move exportCsv into utils (addExportButton stays in map view)
    - Split exportCsv into two functions so that any json can be passed in for reusability
- [x] Add information about selected filters and helper info about what export does
    1) Clicking the 'download' button on the sidebar opens a modal
    2) Modal has intro describing what they are downloading (Each row is one project and all projects are always included, , it includes zone data, the 'matches_filters' is whether it matches the filters currently selected. 'matches_filters' will change each time the data is downloaded, but the rest of the data will be the same.)
    3) The button that actually downloads the .csv file is on the bottom of the modal

    Using the modal gives us a chance to explain what it is they're getting, as well as clearly display the active filters to make sure it's clear how those are connected. We could add an additional 'download filters list' button as well, but first pass having them copy-paste if they want it is probably enough.
- [ ]  List out all the active filters in a copy-pastable format
- [x] Move button to the subheader in bottom half of sidebar

PR #516, opened 2017-09-02
Added filters to export CSV modal and getActiveFilterValues to filters.js for #499

jgordo04:
Added filters to export CSV modal and getActiveFilterValues to filters.js for #499


docs/tool/js/util/filter.js
```diff
@@ -152,18 +154,30 @@ var filterUtil = {
...
             return each.source === msg.split('.')[1];
         });
         filterView.filterInputs[dataChoice.short_name].callback(); // fire callback as if the associated filter has been changed
-    }
+  },
+	getActiveFilterValues: function() {
+	    var allFilterValues = filterUtil.getFilterValues();
+			var currentFilters = {};
+			Object.keys(allFilterValues).forEach(function(key){
+				if ( allFilterValues[key][0] ){
```
NealHumphrey: This isn't actually excluding all unused filters. If a filter is set and then cleared (including some that are "set" on load, i.e. all the ones with chloropleth data), `allFilterValues[key][0]` will exist (true) but it will be an empty array `[]`. This actually means it's an inactive filter and should be excluded from this list. Should add another condition to exclude these false positives.

docs/tool/js/views/map-view.js
```diff
@@ -1019,12 +1019,26 @@
...
+                activeFiltersString += key + ": ";
+                for (var i = 0; i < activeFilters[key].length; i++ ){
+                  activeFiltersString += activeFilters[key][i];
+                  if ( i < activeFilters[key].length - 1 ){
+                    activeFiltersString += ",";
+                  }
+                }
+                activeFiltersString += "\n";
+              });
+              filtersText._groups[0][0].innerText = activeFiltersString;
```
NealHumphrey: It won't be obvious to users what this means: `proj_units_tot: 4,410,true`. Instead, we should use the display_name of the filter (you can get this from the config data stored in `filterView.components` or by iterating over the ui controls themselves via `filterView.filterControls` and looking in the 'component' attribute). 

Also, when it is an array of [min,max,include_nulls] (i.e. date or continuous), we should label them as such. 

I suggest grouping the filters into groups based on `component_type` so that you can make a properly formatted table for each type - 

continuous, date: 3 columns for min, max, include_nulls
location, searchbar, categorical: list of matching categories. 

To prevent future 'missing filters' bugs in case we add new types of categories, use an 'else' at the end to catch anything else and dump it into a string as -is.

NealHumphrey:
@jgordo04 Tables look good! There are a few enhancements/fixes I think should happen, but rather than put them into this PR I've made them into a separate issue:
https://github.com/codefordc/housing-insights/issues/547

I think they should be pretty quick adjustments, you have the bulk of the code in place for this already. 

Thanks!

Issue #514, opened 2017-08-31
Null data breaks some map view functionality


NealHumphrey:
Our newest data update has some null values for fields that we are using in the map view, most notably latitude and longitude but also other geocoded stuff like Ward. We haven't had null data before so haven't properly handled these. 

- Null lat/lon means clicking on an entry from the list of "matching projects" from the list on the right zooms you to the project location off the coast of Africa. 
- Ward-level bar chart of filtered data does not show up because results-view.js summary stats code breaks due to missing ward data. 
- Maybe others. 

Should fix and handle these.

ostermanj:
Adressed in PR #517 

But issues remains: why do projects with know addresses have null lat/lin?

PR #517, opened 2017-09-02
Issue 514: FilteredStats iterates only if ward is not null; flyto on previw clic…

ostermanj:
…k only if ling/lat is not null; alerts otherwise

re: #514 

Makes sure ward !== null before iterating over the filteredStats object.

Disable map flyTo if the selected project lat or lon is null. Shows message instead: 

message
 
Upstream issue, though, seems to remain: why is the location data null for projects with known addresses?

### DIFF
--- a/docs/tool/css/styles.css
+++ b/docs/tool/css/styles.css
@@ -534,9 +534,6 @@ a.active {
     padding: 10px;
     color: rgb(189, 54, 33);
 }
-.no-location-alert.loaded  {
-    color: magenta
-}
 
 #reset-zoom {
     display: none;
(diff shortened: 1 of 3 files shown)


NealHumphrey:
@ostermanj Thanks for these fixes, looks great!

With respect to the upstream issue - the large quantity of null data is something I need to fix; basically we're not finishing geocoding the addresses in the DHCD data set, whereas in the two data sets we had added previously (prescat and dchousing) the latitude/longitude geocoding had already been done for us. 

However, even after that is fixed there will remain some null values. This will occur when we have invalid addresses and therefore cannot successfully geocode the location. There are several cases in the current data where this will happen.

Issue #496, opened 2017-08-26
Additional address format parsing


NealHumphrey:
Moving this into an issue instead of the merged pull request so it's easier to find:

RE: #494 that created the address splitting function

We may end up needing a few more situations - unfortunately there's a bunch more messy formats. However I'd also like to send some of the one-off problem cases back up to the source and see if they'll fix them in their database instead of adding lots of extra logic to deal with it all on our end. We might consider adding a hotfix list for situations like that. Some examples in the DHCD data set that look like they might not be handled in this (albeit already robust) solution:

- Incomplete use of dash address: `1401-09 11TH ST NW`
- Comma separated numbers: `1825,1827,1829,1837,1839,1841,1843,1845,1860,1862,1864 CENTRAL PLACE, NE`
- Mixing of commas and & for numbers: `24, 52, & 230 BATES ST NW`
- Mixing of commas and dash separated numbers with &: `1416-1428,1432,1436 & 1440 R ST NW`
- Different meaning of  &: `HAYES STREET &ANACOSTIA AVENUE, NORTH EAST` (this would be falsely split into two addresses, even though it's actually a street corner that should be one invalid address)

The one of those that I think makes sense to try to add is the comma separated numbers. The other odd ones I might put into a hotfix list that we map from current messy format into a parseable format and put the hotfix mapping as the first step of the get address function. What do you think?

NealHumphrey:
Ha ha so there's another bug in the current address splitting function. Here's two address that were split out in the current code:

NL000382 | 919 Rhode Isl | 294859
NL000382 | Avenue NE |  Null


Found the "and" in `Island` and split the address. 

Is change the 'and' split string to `' and '` enough or do we need something more robust?

NealHumphrey:
FYI I did the `' and '` fix in current PR #494 and that fixed all the examples that occured in the prescat.

NealHumphrey:
@jkwening I think all of this was successfully addressed in #518. Closing.

PR #518, opened 2017-09-04
496 - Additional address format parsing

jkwening:
## Summary

Updated 'get_unique_addresses_from_str()' to handle the following cases noted in #496:
- incomplete use of dash address
- comma separated numbers
- mixing of commas and & for numbers
- mixing of commas and dash separated numbers with &
- different meaning of &
- cleanly handles cases that may be considered invalid address: logs as invalid and returns original string

## Testing
Review cases in unit test, run to confirm all tests pass. Add additional test cases if I missed any edge cases and run unit test.

### DIFF
--- a/python/housinginsights/tools/misc.py
+++ b/python/housinginsights/tools/misc.py
@@ -3,6 +3,7 @@
 from housinginsights.sources.mar import MarApiConn
 import logging
 
+
 def quick_address_cleanup(addr):
      #Used to perform common string replacements in addresses. 
         #The key is original, value is what we want it to become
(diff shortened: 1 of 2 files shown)


NealHumphrey:
@jkwening haven't been able to test this yet - I can either test it tonight and merge, or tomorrow morning at the latest.

NealHumphrey:
@jkwening I ran this code as part of the `get_api_data.py` function using the dhcd data source, which is the one that had many of these problems. I ran into two additional problems that needed to be handled via an invalid address:
- address that ended in semicolon (resulting in an empty second string)
- address that did not have a number at the beginning

I went ahead and fixed these two problems by:
- Adding an `if len(addr_str) > 0:` just inside the multiple_address_results iterator
- Putting this code around the delimiter_stack.pop() line:
```
try:
    delim = delimiter_stack.pop()
  except:
    #this will trigger the invalid address 'else' statement
    delim = ''
```

Hopefully those are good enough solutions - they appear to have allowed the code to run. 

While I did that, I added some enhancements to our error handling in the dhcd sources file where this gets used. All those changes are in this commit: 
https://github.com/codefordc/housing-insights/commit/5bf9502fc29db583f518495a0676e922d046aa52

I'm merging that into dev as I think it's all working properly now.

PR #520, opened 2017-09-05
Issue 469 - Communicating null values

jamiecatania:
@NealHumphrey - A couple questions:

1. It took me a little while to understand how filtering works, but I think I'm understanding the gist of it now. Does this approach make sense for un-toggling the checkbox on initially setting the filter? The toggle is being unchecked as expected, but I'm sure there's a better place to do this check.
2. I haven't seen many other bootstrap components used throughout the app, so wondering if using a bootstrap hover component is sufficient for the hover. The only other place I see popups use mapbox, which I don't believe would be applicable for the sidebar. 

Additional notes:
- Let me know if there's a better approach to implementing the bootstrap component, if needed. It felt awkward dropping jQuery into vanilla js at `docs/tool/js/views/filter-view.js:36`
- I've done very little styling to this hover, as I wanted to get confirmation on the approach first. I'll style it if using the bootstrap component is sufficient.
- Could use some final tooltip text.


NealHumphrey:
@jamiecatania 
1. Yes, the filtering system is a little complex - probably an area that's worth a cliff notes writeup. The way you've implemented it will indeed work the first time the filter is applied. However, if you then clear the filter and reapply it, the checkbox won't uncheck this time. I think that we also want it to occur in this situation. I think you can do this by adding a third conditional to your if statement: `|| filterValues[component.source][0].length == 0` - a cleared filter appears as an empty array instead of [min,max,nullflag]. 

2. This tooltip approach is OK. We have avoided JQuery in general, but we ended up needing it for the semantic-ui searchbar component and couldn't find a dependency-free version. Using it for a tooltip is OK too. Since the tool page load will always be a bit slow due to our API calls, I've tended toward preferring things that keep developers working faster over things that are more efficient on the client end. 

IF you can fix and test that third conditional, style away. I'll think about the best hover over text and can edit it post-merge or will share it here.

jamiecatania:
Cool, thanks for the feedback @NealHumphrey. Good call on the third conditional, I hadn't considered that scenario. Added the third conditional and everything is working as expected locally. Just pushed up that fix, moving on to styling the tooltip window now. Should have that ready shortly.

NealHumphrey:
Hey @jamiecatania on Tuesday evening you said you had the styling stuff ready (almost ready?). Is all your current code pushed up?

I am going to merge this as is, which looks like it contains the third conditional but hasn't edited any of the styling yet. If you push up any changes to styling, be sure to open a new pull request.

Issue #501, opened 2017-08-29
September Master Issue


NealHumphrey:
This is still a work in progress - I need to create some issues for bullet points here, as well as go through all open issues to add any additional ones that need to be addressed before launch. 

Map View
- [ ] URL encoding finalizing (project in preview screen, 'summary' or 'matching' sidebar, fixing color coding on decode)
- [ ] (Jamie G.) Download CSV enhancements (#499)
- [ ] Search and filter by many addresses and TOPA (#500)
- [ ] Enhance project preview (list all subsidies, icons from project view)
- [ ] Change map view to only have 2 colors (match/nomatch)
- [ ] Clear 'Selected Project' if not part of a newly run filter

Project View
- [ ] Revamping project view #471 - need to turn this into multiple issues
- [ ] "Tabulate" as a D3 reusable chart
- [ ] Fix affordable housing nearby issues (#487)

Front end - If time
- [ ] Bar chart improvements (#390)
- [ ] Reduce duplicated geoJSON (#433)

Data Sources (needs both Python and JS)
- [ ] REAC #425 (status: calculated in project table, needs entry in dataChoices.js. Might also need to be added to api?)
- [ ] Nearest metro #424
- [ ] Topa #420

Data Processing 
- [ ] Add nlihc_id using multiple addresses 
    - [ ] TOPA table (refer to #404 for reference)
    - [ ] DC tax
    - [ ] Address
- [ ] Update data documentation for all tables currently in the database
- [ ] Add ANC's to weighting; add ANC to zone facts. 
- [ ] Set up a s3 bucket for uploading PresCat

## General Website Stuff
- [ ] Add "?" links to all data fields linking to the appropriate data documentation page. (#360)ZoneFacts table will provide this for any calculated fields. 
- [ ] Add a landing page
- [ ] Add 3 saved analyses (using the url encoding) and use these to introduce the types of things the tool can do to new users; put these on the landing page. 
- [ ] Add a the navbar from the whole site to the top of the tool
- [ ] Add Google Analytics
- [ ] Auto-population of data fields in Jekyll data documentation #105

ostermanj:
@NealHumphrey I can jump on the finalizing URL encoding stuff. What's the issue with color-coding on decode?

NealHumphrey:
[...]

PR #521, opened 2017-09-06
Minor updates to documentation for select tables in prod db

emkap01:
@NealHumphrey this PR addresses the general 'update data documentation for all tables currently in the database' requirement from Issue #501.  Feel free to hack away at these as you see fit, my intent was just to try and cover some of the basics.

Most of the Markdown files that needed work should be in pretty good shape by this point, so next I will turn my attention to whatever values we may still be missing in the `display_text` key for each dictionary.  Will make sure to deconflict with @louvis.


NealHumphrey:
Cool, merging as-is. We can continue working on the documentation as we go - when we firm up a couple more items in the dedupe process for projects I'll want to add a few edits to make sure that description matches our code process.

PR #523, opened 2017-09-06
Adds nearest_metro_station & bus_routes_nearby to project table.

pfjel7:
- Adds two new entries to meta.json for project table.
- Adds code to _populate_calculated_project_fields() for LoadData.
- Comment explains how bus_routes_nearby still could be updated more efficiently.

### DIFF
--- a/python/housinginsights/ingestion/LoadData.py
+++ b/python/housinginsights/ingestion/LoadData.py
@@ -453,9 +453,10 @@ class LoadData(object):
         conn.execute(stmt)
 
 
-        #########################
+        ########################
         # Sum of tax assessment
-        #########################
+        ########################
+        
         logging.info("  Calculating tax assessment statistics")
 
         stmt= '''
@@ -497,6 +498,7 @@ class LoadData(object):
         # (due to missing addresses in the proj_addre table) we'll be missing the same ones from numerator
         # and denominator so will get realistic average. 
         logging.info("  Calculating TOPA statistics")
+        
         stmt =  """
                 update project
                 set (topa_count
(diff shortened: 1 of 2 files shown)


NealHumphrey:
@pfjel7 thanks for this commit. On a quick read of the code it looks like it should work properly. Unfortunately I won't have a chance to test it until Monday - I have some commitments this afternoon and family in town. 

The improvement that you note in updating the whole table with a bulk insert would be valuable, I think. How fast does the process run on your machine? The row-by-row database queries are one of our biggest bottlenecks in terms of performance. It's certainly not necessary but if you are able to take a crack at it and adjust it without too much effort it would be worth a try. If it ends up being more difficult than expected we can use with the line-by-line query instead. 

Thanks
Neal

pfjel7:
@NealHumphrey, thanks for the comment. Yes, I will work on the bulk update. Already have started thinking about how to do that. Per

pfjel7:
@NealHumphrey  I've now pushed my new commit onto the branch '424-add-nearest-metro . . .'. As I also wrote in a comment before those revisions: 

        The [new] code reduces the time individual database queries
        otherwise would take for record-by-record updates. It gathers 
        the data needed in one initial composite query, uses list comprehension
        to reduce embedded loops, and executes the final update within  
        SQLAlchemy's "executemany" context.

pfjel7:
@NealHumphrey  Used SQLAlchemy's Expression Language to speed up calculation of the tax assessment statistics. Previously took 45 seconds, now 1.5. Helps also that I included a where condition not to join project and dc_tax tables where nlihc_id was missing in the tax table. This revision retains some redundancies, especially in code imports and variable assignments, mainly to make the separate sections for tax assessments and transit stats more easily readable.

NealHumphrey:
@pfjel7 great! I'll test and merge this in this afternoon. I appreciate your work on this!

NealHumphrey:
@pfjel7 I've merged this in. Code looks good, and tax assessment is definitely much faster. I ran this on my machine and using the example we emailed about the NL000180 appraised value came out correct. I'll look into seeing about adding error logging about tables that aren't dropped.

PR #525, opened 2017-09-08
Jo 501 finalize url encoding

ostermanj:
— Fixes color coding not loading on URL decode
— Encodes/decodes navBar.right = buildings status
— Encodes/decodes previewBuilding

Third item got complicated and led to other things:

— checks that project layers have in fact been rendered before initializing the url decoding. That was the source of the timing issues
— marks the previewBuilding dot with a red ring
— changes legend to match that and to remove last-filter designations
— Zooms to selected previewBuilding and disables zoom-to-fit-filter-results functionality when there is a previewBuilding selected. It felt like unwanted behavior to have the map zoom away from a selected project. (TO DO: add a way to unselect the previewBuilding)
— scrolls the matching projects list to show the previewBuilding list item (if it matches filters). That reinforces the connection between the list and the map. Also adds up-triangle icon to indicate that that item is shown in the preview pane.
— scroll happens whether by clicking on the map or by url decode.

I did some pretty good stress testing, selecting lots of filters and a previewBuilding. Seems to be working well. For example: http://127.0.0.1:4000/tool/#/HI/sb=bdng&pb=8d6ec119-9534-4426-9b28-ce6d5e73101c&pa=1-28&pu=4-290&cv=82-430&pov=0.09-0.18&l=Ward_3

### DIFF
--- a/docs/tool/css/styles.css
+++ b/docs/tool/css/styles.css
@@ -436,7 +436,7 @@ Others are working correctly.
     font-size: 1em;
 }
 
-#projects-list>div:hover, #building-view-segments div:hover {
+#projects-list>div:hover, #building-view-segments div:hover, #projects-list>div.projects-list-selected {
     background-color: #ddebff;
 }
 
--- a/docs/tool/css/styles.css
+++ b/docs/tool/css/styles.css
@@ -420,6 +420,7 @@ Others are working correctly.
     font-weight:bold;
     font-size:1.2em;
     cursor: pointer;
+    color: #bd3621;
 }
 #projects-list>div, #building-view-segments div {
     overflow: hidden;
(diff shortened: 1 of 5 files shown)


ostermanj:
re #501

NealHumphrey:
@ostermanj this is great. I agree w/ the choices you made re: highlighting selected building, swapping out the meaning of an outline of the building dots, etc. Approving and merging. 

Note there is one small problem I found that this messed up with the page CSS a bit. On initial page load there is a page scroll bar with a tiny bit of play in it. If you then click to the project page and back, the scroll adds an extra full page of empty space at the bottom of the page. 

There were only a couple CSS changes - were those needed for having the autoscroll work? Or was there any other reason?

PR #527, opened 2017-09-11
Move tool to Jekyll

NealHumphrey:
Moves the tool/index.html content into a toolLayout.html and adds yaml front matter so that the tool is rendered via Jekyl

Fixes various CSS problems due to the conflict caused when the main.css and other sitewide files were introduced

Removes a few console.log statements to speed up performance.

### DIFF
--- a/docs/_includes/header-includes.html
+++ b/docs/_includes/header-includes.html
@@ -8,8 +8,12 @@
   {% endif %}
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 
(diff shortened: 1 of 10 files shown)


NealHumphrey:
@ostermanj I'm merging this so that it's there for the Code for DC hacknight. I think I got all the css conflicts (including a really weird one that was messing up the hover cursor and ability to select data choices on the left side - overlapping padding meant it acted like it was hovering over the data choice above/below). Do keep an eye out for any additional potential conflicts especially between main.css (sitewide) and styles.css (tool specific).

PR #531, opened 2017-09-12
Categorical sorting

NealHumphrey:
All 3 categorical input elements are sorted alphabetically before dropdown list elements are added to the page


NealHumphrey:
image

Issue #535, opened 2017-09-12
Sidebar doesn't fully reset URL encoding


NealHumphrey:
When the URL decoding encounters an encoded tool url without any parameters, it is not able to remove the 'Loading saved state' message:

URL format: `http://127.0.0.1:4000/tool/#/HI/`

Error message:
![image](https://user-images.githubusercontent.com/5642566/30336221-82e76dcc-97b2-11e7-8e41-519bb2bf9fb5.png)

How to recreate the error:
- load the /tool page
- Select the 'matching projects' sidebar, and then reselect the 'summary' tab. 
- Refresh the page

Clearing the active filters ('x' on the pillboxes) appears to remove the /HI/ portion of the url when no parameters remain; however, the sidebar doesn't behave this way. It's probably best to fix it on the decode, but could also fix the sidebar encoding to similarly remove the /HI/ portion when no encoding remains.

NealHumphrey:
@ostermanj can you take this on? Should be quick fix.

ostermanj:
Yeah, I'll take a look.

ostermanj:
Closing

PR #539, opened 2017-09-12
Fixes path #/HI problem

ostermanj:
Fixes the encoding mistake leading to path ending in #/HI. Also provides decoding backstop just in case that crops up again.

RE #535


ostermanj:
I'll add the mm/dd/yyyy fix to the date filters on top of this PR because it needs the #/HI fix. Coming soon.

NealHumphrey:
@ostermanj I'm approving this PR now. If you push the date filters fix, can you open a new PR?

PR #541, opened 2017-09-12
updated dataChoices.js with 'sourcetable' field for each dictionary

emkap01:
Per discussion this evening, I updated dataChoices.js with a `sourcetable` key per dictionary, and then populated the value for each key with the corresponding table from which that value is derived.


NealHumphrey:
Great, thanks @emkap01

PR #543, opened 2017-09-13
Adds streetview image to project view

ostermanj:
Uses address string. Some may come up without image available. Lat/long is possible, but then you have to set heading and pitch of the camera.

screen shot 2017-09-12 at 8 26 54 pm


ostermanj:
Ah, I branched this from an unmerged branch. Will delete this PR and redo from clean branch.

PR #546, opened 2017-09-13
updated meta.json display_text fields

louvis:
I went through and added some more display_text values and tried to standardize what I had entered along with existing entries. The one table I didn't finish that I worked on was dc_tax because there are some fields that I can't find a good reference for.

### DIFF
--- a/python/scripts/meta.json
+++ b/python/scripts/meta.json
@@ -5,7 +5,7 @@
     "fields": [
       {
         "display_name": "NLIHC ID",
-        "display_text": "Unique identifier of the project as found in the project table.",
+        "display_text": "Preservation Catalog ID",
         "required_in_source": true,
         "source_name": "Nlihc_id",
         "sql_name": "nlihc_id",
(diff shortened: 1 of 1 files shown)


NealHumphrey:
Looks good. 

@louvis on the tax one, I found this "dataset" on opendata.dc.gov - data dictionary for the integrated tax system: http://opendata.dc.gov/datasets/integrated-tax-system-data-dictionary/data

emkap01:
I filled in some (but not all) of the remaining missing values based on the data dictionary Neal found.  In several cases the label of a field in that dictionary was not an exact match for the way the same field is labeled in our `dc_tax` table, however.  For example:

The field `assessor_name` appears labeled as `ASR_NAME` in the data dictionary.  But based on the field description in the dictionary (i.e., "Name of property assessor"), it's clear they are the same field.

I double checked both the field description in the data dictionary as well as the actual values within our table to make sure that I was matching the correct description to the correct field.  

Since I was erring on the side of caution, I did not add a value to the `display_text` key if the match seemed ambiguous or unclear.

Issue #537, opened 2017-09-12
dateFilterControl should be month first


NealHumphrey:
The current date filter control is set up to have text fields in the format [day] / [month] / [year]. This is likely not intuitive to most US users who will expect mm/dd/yyyy. I did not realize it was day first until adding a new field that had days outside of the 1-12 range. 

Change order of input and make sure filter applies them properly. Add key somewhere (maybe small gray text below the fields) to indicate the proper order.

ostermanj:
Will address this on top of PR #539 because it needs the #/HI path fix to work.

PR #550, opened 2017-09-15
Date input validation / alerts;  decode url fixes

ostermanj:
* Changes date filter  format to mm/dd/yyyy
* Adds note to that effect
* Validates that dates put in the text boxes are valid and within range
* Alerts user if an input is invalid

image
image
image

Also:
URL decoding was sometimes happening before the `filterControlsDict` was defined. 
* Adds gate function to check multiple states before initializing URL decoding.

### DIFF
--- a/docs/tool/js/core/router.js
+++ b/docs/tool/js/core/router.js
@@ -3,7 +3,16 @@
 var router = {
     isFilterInitialized: false,
     stateObj: {},
-    initFilters: function(msg,data){
+    initGate: function(){
+        if ( getState().initialProjectsRendered && getState().filterViewLoaded ) { // each stateChange triggers the gate function
+                                                                                   // but both need to be true before calling
+                                                                                   // initFilters. fixes bug whereby url decode
+                                                                                   // was running before the filterControlsDict
+                                                                                   // was defined
+            router.initFilters();
+        }
+    },
+    initFilters: function(){
         if ( router.isFilterInitialized ) return;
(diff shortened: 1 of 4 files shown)


ostermanj:
re: #537

docs/tool/js/core/router.js
```diff
@@ -3,7 +3,16 @@
 var router = {
     isFilterInitialized: false,
     stateObj: {},
-    initFilters: function(msg,data){
+    initGate: function(){
+        if ( getState().initialProjectsRendered && getState().filterViewLoaded ) { // each stateChange triggers the gate function
```
NealHumphrey: This is great. I was concerned about this when I was working on the "loading" screen because that projectsRendered event was just based on the slowness of mapbox not the sequencing of all the stuff needing to occur.

 We should think about whether it makes sense to mess around a bit more with the sequence and gate functionality so that there can be one master 'page done loading' event that is triggered - when router is needed, it's after router is done loading, but when router isn't needed it is this same gate conditions of what you have put here. Then the "loading" buttons in the modal can be based on that 'loading done' state event rather than the filterviewLoaded that I'm using now.

ostermanj:
Add some functionality and fixes to the date validation, extends much of it to continuous filters. User input less than actual min in data is coerced to the min. Same for max. Enforces min input can't be more than max. Clears error markers when clear pillbox (all or single) is clicked.

PR #552, opened 2017-09-15
Clean up logging

NealHumphrey:
-Removes the logging level from the init params of the logger, resulting in the use of the default. This lets us specify the logging level in just one place. Set new level to info instead of debug.

- Remove all references to 'logging' so that a second stream handler is not created, making the logging easier to read.
- Adds a counter to the cleaner since debug level messages are no longer shown.


NealHumphrey:
@speedyturkey - I found that the problem with the really muddy console output was the mixing of your HILogger class and plain ol 'logging.info' statements. If you use a logging statement anywhere in the code, it automatically adds a new streamhandler that uses a different format. This PR replaces everything so we're back to a clean output instead of having duplicated statements. 

image

PR #553, opened 2017-09-15
Error handling and moving drop data until after cleaning

NealHumphrey:
This takes a few steps towards making our data loading process more robust

- Adds a "debug" mode to the LoadData class (in the __init__ params). If debug mode is true, errors are raised when they occur.
- Wraps the row-level cleaning in a try/except function; if it fails, write an error to the logs, and if in debug mode reraise the error
- Stronger warning (logger.error) when data can't be deleted from the database, and raises the exception if in debug mode. @pfjel7 this would help alert you to the problem that you encountered of an open connection preventing dropping of the table and ending up with duplicate rows. Not complete though - we should work more on this so that it prevents the new data from coming in if the data couldn't be dropped. 
- Moves the deletion of data to occur after data cleaning so that it is just before data loading, instead of having time inbetween when no data  is in the database.
- Eliminates the requirement for the sql manifest row to have status !=loaded. This is an old requirement from when the unique_data_id was meant to correspond to one specific file. In our current configuration, the unique_data_id is constant across all copies of the same data set, so even if it is loaded if we are running the function it means we want to overwrite that dataset. This requirement was essentially unused due to the fact that we deleted the data from the database before checking the sql manifest, so it was impossible to trigger in the previous code but moving the timing of data deletion has changed that

### DIFF
--- a/python/housinginsights/ingestion/DataReader.py
+++ b/python/housinginsights/ingestion/DataReader.py
@@ -369,29 +369,18 @@ class DataReader(HIReader):
 
     def _check_include_flag(self, sql_manifest_row):
         """
-        Internal function compares manifest from the csv to manifest in the
-        database. If the manifest says the file should be used ("use") AND the
-        file is not already loaded into the database (as indicated by the
-        matching sql_manifest_row), the file will be added.
+        Checks to make sure the include_flag matches requirements for loading the data
 
-        The sql object in charge of getting the sql_manifest_row and writing
-        new sql_manifest_row elements to the database is in charge of making
-        sure that the sql_manifest_row['status'] field can be trusted as a true
-        representation of what is in the database currently.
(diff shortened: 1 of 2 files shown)


NealHumphrey:
@jkwening I think that all this stuff is good to go, but would be worth a second set of eyes especially on the moving of the timing of dropping the existing data. Note, there's definitely more that needs to be done on making this process more robust and handling all the errors appropriately so that the admin actually realizes that something wrong has occured when the server sends them the weekly email. Baby steps.

NealHumphrey:
@jkwening I'm going to merge this in b/c I need it for tomorrow's deploy and for testing of the ingestion code on the server. But, please do come back and take a look at this, at least a skim through the code so you know what the types of changes I made were surrounding moving the delete rows function.

PR #555, opened 2017-09-16
Display a table of subsidies in the project preview pane.

ptgott:
Addresses #476 and #501 

I haven't attempted to style the table.

There's also a possible issue with the building information overflowing its container and requiring you to scroll (if we're not keen on too many scroll bars)

### DIFF
--- a/docs/tool/js/views/map-view.js
+++ b/docs/tool/js/views/map-view.js
@@ -846,7 +846,7 @@
                     }, 500)
                 }
         },  
-
+        
         showProjectPreview: function(msg, data) {
             var projectData = data[0];
             if ( projectData ) {
@@ -917,8 +917,7 @@
                     current.append('br')
 
                     //Add a definition list of property: value
-                    var previewFields =     ['proj_units_assist_max', 'proj_units_tot','subsidy_end_first',
-                                            'subsidy_end_last']
+                    var previewFields =     ['proj_units_assist_max', 'proj_units_tot']
 
                     var dl = current.append('dl')
                             .classed("properties-list",true)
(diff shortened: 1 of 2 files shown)


docs/tool/js/views/map-view.js
```diff
@@ -932,6 +931,32 @@
...
+                                .append('div').attr('id', subsidyTableId);
+                                                
+                            new D3Table('#' + subsidyTableId)
+                                .data(data.items)
+                                .columns([
+                                    {field:'poa_end', label:'Scheduled End Date', class:'value', html: function(d){return d}},
+                                    {field:'poa_start', label:'Start Date', class:'value', html: function(d){return d}},
+                                    {field:'program', label:'Program', class:'value', html: function(d){return d}}
+                                ])
+                                .create();
```
NealHumphrey: If you click the same property multiple times, the `create()` function is run multiple times, adding additional tables to the div. 

The best way to fix this is probably to wrap the whole `showProjectPreview` function in an if statement that only runs the update if the newly selected project is not the same as the previous project. While we could do it at the table level, there could be other things added int he future that have the same problem and this would avoid them. 

OR we could change how the behavior works so that the existing project preview is ALWAYS faded out before bringing in the new content, even if you've re-selected the same building. That would also fix a usability problem of people who wondered why nothing happened when they clicked a project if it was the one already selected. The data binding around row 865 is what is causing it to change only if it is a new selection, so that solution starts there.

docs/tool/js/views/map-view.js
```diff
@@ -932,6 +931,32 @@
...
+                                .attr('id', subsidyContainerId)
+                                .append('p')
+                                .text('Subsidies for this project');
+                            d3.select('#' + subsidyContainerId)
+                                .append('div').attr('id', subsidyTableId);
+                                                
+                            new D3Table('#' + subsidyTableId)
+                                .data(data.items)
+                                .columns([
+                                    {field:'poa_end', label:'Scheduled End Date', class:'value', html: function(d){return d}},
```
NealHumphrey: Let's change these labels to be shorter given the space constraints. How about just "Start" and "End". I also think we should maybe only display the start/end year, to save space - users can see the full date by clicking through to the proejct page. Finally, do put a "-" or similar null character marker if there is missing data like date.

NealHumphrey:
Edit - when I refreshed the page, the consoler error bug I noted has dissappeared. Maybe there was something conflicting stuck in my cache.

PR #556, opened 2017-09-18
Passes sums of residential_units (by zone) to other LoadData methods.

pfjel7:
- Calculates residential_units from mar table.
- Records sums in zone_facts table.
- Removes zone_housingunits... row from manifest.csv.
- Marks cama.py as deprecated.
- Removes unused _add_building_permits_rate().

### DIFF
--- a/python/housinginsights/ingestion/LoadData.py
+++ b/python/housinginsights/ingestion/LoadData.py
@@ -643,9 +643,10 @@ class LoadData(object):
             sql_interface.create_table(db_conn=conn, table='zone_facts')
 
         # populate table with calculated fields values
-        self._populate_zone_facts_table()
+        res_units_by_zone_type = self._get_residential_units()
+        self._populate_zone_facts_table(res_units_by_zone_type)
 
-    def _populate_zone_facts_table(self):
+    def _populate_zone_facts_table(self,res_units_by_zone_type):
         """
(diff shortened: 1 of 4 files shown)


NealHumphrey:
@pfjel7 you're right that the `_populate_zone_facts` method is a bit confusing, and I see why you separated this out. For some background, this code was originally in the api itself and would calculate summary statistics on the fly based on user input of zone type and desired field. So a lot of this code is not well designed for its current use. Overall it would be good to refactor it to be cleaner considering the current purpose of populating the table.

I actually recommend a slight modification to your approach. I would suggest putting all the code that fills in the residential_units into `_populate_zone_facts, but put it above the `for field in summarize_obs_field` loop. You can leave the separate `_get_residential_units()` function where it is, but call it as the first line of `_populate..` rather than passing the residential units as an argument to populate. That way you don't need to force-fit the residential units into the for loop. The for loop is for applying weighting factors and doing normalization, but residential units don't need this since you get everything you need from the _get_residential_units method directly. This also makes sure that the residential units is run before the code that depends on it (building permits). 

I'm merging this in now and I think it is fine to leave this as is, especially considering we should really refactor the whole thing to be cleaner - but that should happen after the launch since it's just paying down our tech debt not getting us a better working tool for launch day. If you think it would be easier to do a quick version of the refactor now while it's fresh in your mind, go for it.

PR #562, opened 2017-09-19
WIP: Issue-471: Project view Icons

jamiecatania:


jamiecatania:
On the TODO list:

- [ ] Finish the subsidized unit chart
- [ ] Add icons for ownership type
- [ ] Cleanup styling/move style out of d3 `.attr()` into css

docs/tool/js/views/project-view.js
```diff
@@ -222,25 +222,29 @@ var projectView = {
...
+                .columns([
+                    {field:'title', label:'Title', class:'title', html: function(d){return d}},
+                    'value'])
+                .hideTitle(true)
+                .create();
+
+            var projLongitude = full_project_data['longitude'];
+            var projLatitude = full_project_data['latitude'];
+
+            d3.select('#project-location-map').attr('src', 'https://api.mapbox.com/styles/v1/mapbox/light-v9/static/pin-s-star+bd3633(' + projLongitude + ',' + projLatitude + ')/-77.0369,38.9072,8.3/124x124?access_token=' + mapboxgl.accessToken + '&attribution=false&logo=false');
```
jamiecatania: I couldn't figure out how to implement a simple elliptical marker, so this is using the pin marker. I tried setting up a simple GeoJSON to see if it would provide more options for the marker style, but no luck.

NealHumphrey: Pin marker looks great - we'll go with that! 

I might consider making the pin slightly smaller if that's easy, but also fine as is.

NealHumphrey:
@jamiecatania this looks good. I'm going to approve this part so that we get it in there - you can do a separate PR with the additional changes for the remaining icons/sections. 

Thanks!

PR #570, opened 2017-09-19
WIP fixing DHCD data download

NealHumphrey:
Fixes the issue that the "dhcd_dfd_projects" table, which contains loan information, was returning empty. This was due to an error returned from quickbase that the query was too large. Fixed by only requesting default fields for now, and made note for future users that all fields could be obtained by chunking the request.

Also started making progress towards the reporting / QA task for verifying we are properly matching records. Added a 'found_via' parameter to the _get_nlihc_id_from_db method, telling how the element was identified, as well as the list of all the mar_ids that it found using the addresses. Set up useless function that when finished can use these mar_ids to geocode the project so that it doesn't need to be done during hte cleaning step.

TODO:

- map the fields of the dhcd_project table to the subsidy table so that we can use the data
- finish the _geocode_if_needed function; this will mean it doesn't need to occur during cleaning. 
    - Modify append_addresses so that it returns the in-memory copy of what it writes to teh file
    - move _geocode_if_needed below append_addresses, and use the resulting mar_ids to geocode the project.

### DIFF
--- a/python/housinginsights/sources/base.py
+++ b/python/housinginsights/sources/base.py
@@ -107,7 +107,7 @@ class BaseApiConn(object):
         else:
             url = urlpath
 
-        logger.debug("Requested URL %s with params %s", url, params)
+        logger.info("Requested URL %s with params %s", url, params)
         return self.session.get(url, params=params, proxies=self.proxies, **kwargs)
 
     def post(self, urlpath, data=None, **kwargs):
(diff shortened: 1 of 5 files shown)


jkwening:
The work in this PR has been merged into #573. @NealHumphrey I assume this PR can be closed since your work will continue from what's in my PR?

PR #573, opened 2017-09-20
505 base proj apiconn

jkwening:
## What's In This
Generate _project.csv and _addre.csv according to these rules:
- don't add a new row if there is a duplicate project with different unique_data_id
- add new row if there is a duplicate project with same unique_data_id
- add a new row if no matching project is found

Uses HILogger to log results from determining whether to add new row or
not. Merged in 136-fix-dhcd-loan-table-download branch. (#570) 

## Testing
- run get_api_data.py on "DCHousing" and "dhcd" modules and verify that appropriate csv files are generated. The resulting files will be unique projects for the respective raw data vs prescat projects.
- update manifest accordingly, load dchousing_addre into proj_addre, rename current _addre.csv files, and then re-run get_api_data.py on "DCHousing" and "dhcd". The resulting dchousing_addre.csv should be same as the original and "dhcd" should be shorter to account for duplicates across dchousing and dhcd.


python/housinginsights/sources/base_project.py
```diff
@@ -271,64 +249,112 @@ def create_project_subsidy_csv(self, uid, project_fields_map,
...
                         # be tricky since we will want duplicates from the same source that
-                        # have been loaded into the database previously. 
-                        self._append_addresses(addresses=addresses, nlihc_id = nlihc_id, addre_writer = addre_writer)            
+                        # have been loaded into the database previously.
+
+                        # pass '' if address missing
+                        if not addresses:
+                            addresses = ['']
+                        # TODO: probably not a good idea to add with missing
+                        # TODO: address metadata - we should consider to mar
```
NealHumphrey: Can you explain what you mean by this comment?

jkwening: Specifically for dchousing data we have projects that have mar_id metadata but no valid address - this is a known issue. I was just jotting a quick note that we should consider doing the geocode clean up sooner than later because I had to add code to handle the missing address metadata so it didn't break what I was doing..

Based on our conversations yesterday, the TODO can be removed since we seem to be both on the same page regarding this general issue and discussed briefly about how to best handle geocode clean up process.

PR #576, opened 2017-09-21
505 base proj apiconn

NealHumphrey:
Additional changes to get the weekly update job working better
-Added missing tables
- Lat/Lon now added during cleaning if needed
- Added an option to skip recalculation, so that we don't recalcualte the calculated fields in the middle of the process of the weekly update
- added an option to manually drop all the tables first so that we can partially rebuild the database (drop all tables but then use the `--update-only` flag to only load a limited set of tables)
- Used all these flags to allow a version of the weekly_update that starts by dropping all tables. This would not preserve unique uuid4 that are generated, but it allows us to do a fresh check for the existence of duplicates beween prescat and the other project table sources. 
- added a couple error catching try/except blocks to the recalculation routines and moved recalculation into its own function.


NealHumphrey:
Note this is the version that I've used for the live database update on 9/20/2017

PR #591, opened 2017-09-30
583 - made sure mar id is returned as str type and not as integer

jkwening:
## What's In This
In misc.py check_mar_for_address() we were return the address id result from mar api call as integer and using that for dedupe process for *_addre.csv. This was primarily occurring for dhcd raw data since it isn't processed with any address id metadata. Simply cast it as str type before returning the result.

## Test
Run get_api_data on dhcd unique data id and verify that all mar ids are string. Example is 1354 Euclid St.

## TODO
Since we're already calling mar api to get mar id for missing projects, we should go ahead and store that value in the output.csv (_project.csv & _addre.csv) instead of making another api call when we do geocode clean up in the cleaner code. I'll create a ticket to expand on this further.

### DIFF
--- a/python/housinginsights/tools/misc.py
+++ b/python/housinginsights/tools/misc.py
@@ -5,6 +5,7 @@ from housinginsights.sources.mar import MarApiConn
 from housinginsights.tools.logger import HILogger
 logger = HILogger(name=__file__, logfile="sources.log")
 
+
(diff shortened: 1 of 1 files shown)


NealHumphrey:
@jkwening Thanks for this! I'll re-run the load code tomorrow w/ this and see how it impacts our matching.

PR #609, opened 2017-10-30
Expands on the instructions for deploying the app to the AWS server.

NealHumphrey:


NealHumphrey:
@jkwening Here's my updated instructions for deploying

PR #618, opened 2019-10-09
Large Refactor

ndtallant:


NealHumphrey:
Yowzer, this is a big one. It's kinda hard to do much review since the renaming of folders means the diff shows up just as a bunch of deleted and added files, so I can't tell what's been edited and what hasn't - typically I'd recommend this kind of refactor occur in two pull requests, one where you rename folders/files and then a separate one where you edit those files. 

I'll go ahead and approve assuming you've done your due diligence. But there is one change you have to make before you merge this - you'll need to rename the `docs` folder back to what it was. This is a requirement of Github Pages, which is what we use to host the housinginsights.org website: https://help.github.com/en/articles/configuring-a-publishing-source-for-your-github-pages-site

NealHumphrey:
Also, you'll need to then merge `dev` into `master` so that this actually gets deployed - we had the dev branch as default when we had lots of contributors because it made it easier to create actual releases while bringing in lots of changes.

